PR #3, opened 2016-01-07
list / search locations

floppyears:
These commits allow the user to use full text search for locations and filter locations by campus or type.

### DIFF
--- a/src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
+++ b/src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
@@ -3,6 +3,7 @@ package edu.oregonstate.mist.locations.frontend.resources
 import com.codahale.metrics.annotation.Timed
 import com.fasterxml.jackson.databind.ObjectMapper
 import edu.oregonstate.mist.api.AuthenticatedUser
+import edu.oregonstate.mist.api.Resource
 import io.dropwizard.auth.Auth
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
@@ -12,7 +13,6 @@ import javax.ws.rs.Path
 import javax.ws.rs.PathParam
 import javax.ws.rs.Produces
 import javax.ws.rs.QueryParam
-import javax.ws.rs.WebApplicationException
 import javax.ws.rs.core.Context
 import javax.ws.rs.core.MediaType
 import javax.ws.rs.core.MultivaluedMap
@@ -22,7 +22,7 @@ import java.util.regex.Pattern
 
 @Path("/locations")
 @Produces(MediaType.APPLICATION_JSON)
-class LocationResource {
+class LocationResource extends Resource {
     private static final Logger LOGGER = LoggerFactory.getLogger(LocationResource.class)
     public static final ArrayList<String> ALLOWED_CAMPUSES = ["corvallis"]
     public static final ArrayList<String> ALLOWED_TYPES = ["building", "dining"]
@@ -67,34 +67,31 @@ class LocationResource {
     @Timed
     Response list(@QueryParam('q') String q, @QueryParam('campus') String campus, @QueryParam('type') String type,
                   @Auth AuthenticatedUser authenticatedUser) {
-
         def pageNumber = getPageNumber()
         def pageSize = getPageSize()
         def trimmedQ = sanitize(q?.trim())
         def trimmedCampus = sanitize(campus?.trim())
         def trimmedType = sanitize(type?.trim())
 
-        if (trimmedCampus && !ALLOWED_CAMPUSES.contains(trimmedCampus)) {
-            throw new WebApplicationException(Response.Status.NOT_FOUND)
+        // validate filtering parameters
+        def invalidCampus = trimmedCampus && !ALLOWED_CAMPUSES.contains(trimmedCampus)
+        def invalidType = trimmedType && !ALLOWED_TYPES.contains(trimmedType)
+        if (invalidCampus || invalidType) {
+            return notFound().build()
         }
 
-        if (trimmedType && !ALLOWED_TYPES.contains(trimmedType)) {
-            throw new WebApplicationException(Response.Status.NOT_FOUND)
-        }
-
+        // generate ES query to search for locations
         def query = [ "match_all": [:] ]
         def esFullUrl = getESFullUrl()
-
         def esQuery = getESSearchQuery(trimmedCampus, trimmedType, trimmedQ, pageNumber, pageSize, query)
         ObjectMapper mapper = new ObjectMapper(); // can reuse, share globally
         String esQueryJson = mapper.writeValueAsString(esQuery)
 
+        // get data from ES
         def url = new URL(esFullUrl + "/_search")
         URLConnection connection = postRequest(url, esQueryJson)
 
-        def esReponse = connection.content.text
-
-        Response.ok(esReponse).build()
+        ok(connection.content.text).build()
     }
 
     @GET
@@ -106,10 +103,10 @@ class LocationResource {
         String esResponse = new URL(esFullUrl + "/${id}/_source").getText()
 
         if (!esResponse) {
-            throw new WebApplicationException(Response.Status.NOT_FOUND)
+            return notFound().build()
         }
 
-        Response.ok(esResponse).build()
+        ok(esResponse).build()
     }
 
     /**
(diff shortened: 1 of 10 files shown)


src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
```diff
@@ -0,0 +1,261 @@
...
+     * Default page number used in pagination
+     * @todo: should this be coming from skeleton or configuration file?
+     */
+    public static final Integer DEFAULT_PAGE_NUMBER = 1
+
+    /**
+     * Default page size used in pagination
+     * @todo: should this be coming from skeleton or configuration file?
+     */
+    public static final Integer DEFAULT_PAGE_SIZE = 10
```
ghost, line 41: if you want to be able to set the paging parameters at execution time, it would make sense to read from the configuration file. if you do not expect the paging behavior to change between executions, either define the parameters here or in a different file in `edu.oregonstate.mist.locations`.

i would not recommend modifying the skeleton until this api has implemented one of those options. note that it may not be trivial to implement paging functionality in the skeleton due to the differences between various apis.

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
```diff
@@ -0,0 +1,261 @@
...
+
+    @Context
+    UriInfo uriInfo
```
ghost, line 46: i believe lines 42-43 are unnecessary

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
```diff
@@ -0,0 +1,261 @@
...
+        def trimmedCampus = sanitize(campus?.trim())
+        def trimmedType = sanitize(type?.trim())
+
+        if (trimmedCampus && !ALLOWED_CAMPUSES.contains(trimmedCampus)) {
+            throw new WebApplicationException(Response.Status.NOT_FOUND)
+        }
+
+        if (trimmedType && !ALLOWED_TYPES.contains(trimmedType)) {
+            throw new WebApplicationException(Response.Status.NOT_FOUND)
+        }
```
ghost, line 82: see comment in previous pull request regarding `edu.oregonstate.mist.api.Resource.notFound()`

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
```diff
@@ -0,0 +1,261 @@
...
+        ObjectMapper mapper = new ObjectMapper(); // can reuse, share globally
+        String esQueryJson = mapper.writeValueAsString(esQuery)
+
+        def url = new URL(esFullUrl + "/_search")
+        URLConnection connection = postRequest(url, esQueryJson)
+
+        def esReponse = connection.content.text
+
+        Response.ok(esReponse).build()
+    }
```
ghost, line 86: consider refactoring the elasticsearch connection methods to a separate class.

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
```diff
@@ -0,0 +1,261 @@
...
+    /**
+     * Sanitizes the search query string by replacing illegal characters with spaces.
+     *
+     * @param searchQuery
+     * @return sanitized search query
+     */
+    private static String sanitize(String searchQuery) {
+        if (!searchQuery) {
+            return ""
+        }
```
ghost, line 111: see `directory-api` for [an equivalent way of doing this](https://github.com/osu-mist/directory-api/blob/e54c37030e81f0b1402e3bf3eac70f0ef11dda3f/src/main/groovy/edu/oregonstate/mist/directoryapi/DirectoryEntityDAO.groovy#L202-L212) which returns `null` instead of `""`

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
```diff
@@ -0,0 +1,261 @@
...
+
+            if (entry.key.substring(0, a).equals(key)) {
+                if (entry.key.substring(a + 1, b).equals(index)) {
+                    return entry.value?.get(0)
+                }
+            }
+        }
+
+        null
+    }
```
ghost, line 146: it might be worth implementing this using regular expressions. something like `\[(.*)\]` would enable you to make use of the parenthesized group.

also consider whether this method should be public.

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
```diff
@@ -0,0 +1,261 @@
...
+        def pageNumber = getArrayParameter("page", "number", uriInfo.getQueryParameters())
+        if (!pageNumber || !pageNumber.isInteger()) {
+            return DEFAULT_PAGE_NUMBER
+        }
+
+        pageNumber.toInteger()
+    }
+
+    /**
+     * Returns the
```
ghost: Returns the...

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
floppyears, line 41: My focus was to implement the locations API without spending too much time on getting to the code to be close to perfect. What we'll probably do is add a method getPageSize() and getPageNumber() to the skeleton Resource. The Skeleton resource would take care of using the following technique to identify the values:
1. If page[variable] is in the URL and valid - return it else
2. If configuration.yaml specifies page[variable] value - return it
3. Use a constant such as DEFAULT_PAGE_SIZE or DEFAULT_PAGE_NUMBER specified in the Resource skeleton.

floppyears, line 41: This is something we'll probably focus on during a future sprint since it is code refactoring / clean up and not affecting functionality.

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
floppyears, line 46: This variable is used to figure out pagination. Other languages such as: php, ruby and python allow people to specify arrays in the urls and it converts them for you out of the box. Java doesn't do that so I had to use the uriInfo to inspect the url :(

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
floppyears, line 82: agreed

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
floppyears, line 86: You are right. This class is doing too much. I wanted to see what the final result was like before refactoring things out.

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
floppyears, line 111: Good catch. This is a bad habit of java and trying to avoid NPEs :(

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
floppyears, line 146: I'll leave this as is for now. When this method is moved to the skeleton, we can refactor it to use a regex. This method is used to by pagination (to get value of: page[number], page[size])

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
ghost, line 46: wow, now i finally know what that code does. i've seen it all over in dropwizard but it was never necessary for the applications i worked on.

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
ghost, line 41: that sounds like a good approach

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
ghost, line 146: certainly. although a compiled regex would have better performance, it makes sense to use something that works.

ghost:
i have not verified the behavior, but the commits look good to me. if it functions as desired, merge at your convenience.

PR #4, opened 2016-01-12
Feature/co 270 minor fixes

floppyears:


src/main/groovy/edu/oregonstate/mist/locations/frontend/jsonapi/ResourceObject.groovy
```diff
@@ -0,0 +1,8 @@
+package edu.oregonstate.mist.locations.frontend.jsonapi
+
+class ResourceObject {
+    String id
+    String type
+    //@todo: how do we want to handle this???
+    def attributes
```
ghost, line 6: how about `Object attributes`?

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
```diff
@@ -82,21 +84,109 @@ class LocationResource extends Resource {
         }
 
         String result = locationDAO.search(trimmedQ, trimmedCampus, trimmedType, pageNumber, pageSize)
-        ok(result).build()
+
+        ResultObject resultObject = new ResultObject()
+        resultObject.data = []
```
ghost, line 89: why not initialize `ResultObject.data` in the class definition?

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
```diff
@@ -82,21 +84,109 @@ class LocationResource extends Resource {
...
+                                    ResultObject resultObject) {
+        def totalHits = topLevelHits.get("total").asInt()
+        // If no results were found, no need to add links
+        if (!totalHits) {
+            return
+        }
+
+        Integer pageNumber = getPageNumber()
+        Integer pageSize = getPageSize()
+        def map = ["q": q, "type": type, "campus": campus, "pageSize": pageSize, "pageNumber": pageNumber]
```
ghost: consider indenting for readability. also consider more descriptive variable name.

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
```diff
@@ -82,21 +84,109 @@ class LocationResource extends Resource {
...
+        int lastPage = Math.ceil(totalHits / pageSize)
```
ghost, line 138: verify that integer division behaves as intended. if `5 / 2 == 2`, then rounding will have no effect and a page will be missing.

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
```diff
@@ -82,21 +84,109 @@ class LocationResource extends Resource {
...
+        def uriAndPath = uriInfo.getBaseUri().toString() + uriInfo.getPath()
+        def nonNullParams = params.clone()
+        // convert pageVariable to page[variable]
+        nonNullParams["page[number]"] = nonNullParams['pageNumber']
+        nonNullParams["page[size]"] = nonNullParams['pageSize']
+        nonNullParams.remove('pageSize')
+        nonNullParams.remove('pageNumber')
+
+        // remove empty GET parameters
+        nonNullParams = nonNullParams.findAll { it.value } .collect { k, v -> "$k=$v" }
```
ghost, line 177: a loop would probably be more efficient, though the code would be less dense.

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
ghost, line 138: modifying this line may require change or removal of lines 139-151

src/test/groovy/edu/oregonstate/mist/locations/frontend/LocationDAOTest.groovy
```diff
@@ -0,0 +1,84 @@
...
+
+    private ObjectMapper mapper = new ObjectMapper()
+
+    private final String SINGLE_RESOURCE_ID = "831ed9ce6311601afba57934adea7a8e"
+
+    @ClassRule
+    public static final DropwizardAppRule<LocationsFrontendConfiguration> APPLICATION =
+            new DropwizardAppRule<LocationsFrontendConfiguration>(
+                    LocationsFrontEndApplication.class,
+                    new File("configuration.yaml").absolutePath)
```
ghost, line 22: did this line solve the issue with `configuration.yaml` in `src/main/resources`?

swagger.yaml
```diff
@@ -1,83 +1,224 @@
...
+      type:
+        type: string
+      open_hours:
+        type: array
+        items:
+          $ref: "#/definitions/DayHours"
+
+  DayHours:
+    properties:
+      day:
```
ghost, line 185: [swagger editor](http://editor.swagger.io/#/) gives the error `Data does not match any schemas from 'anyOf'`

swagger.yaml
```diff
@@ -1,83 +1,224 @@
...
+          name: campus
+          description: "Name of the campus. Valid options are: corvallis, bend, newport or extension"
+          required: false
+          type: string
+
+      # Expected responses for this operation:
       responses:
-        200:
-          description: A short message
+        # Response code
```
ghost, line 50: does this resource return any status other than "ok" and "internal server error", such as "bad request"?

src/main/groovy/edu/oregonstate/mist/locations/frontend/db/LocationDAO.groovy
```diff
@@ -46,7 +46,7 @@ class LocationDAO {
      */
     String getById(String id) {
         try {
-            return "${ESFullUrl}/${id}/_source".toURL().text
+            return "${ESFullUrl}/${id?.toLowerCase()}/_source".toURL().text
```
ghost, line 49: will the url be correct if `id` is `null`?

ghost, line 49: why convert `String` to `URL` to `String`?

swagger.yaml
```diff
@@ -1,83 +1,224 @@
...
+        - in: query
+          name: dining
+          description: "Specify 1 for locations that are dining areas"
+          required: false
+          type: string
+        - in: query
+          name: is_open_dining
+          description: "Specify 1 for dining locations that are currently open"
+          required: false
+          type: string
```
ghost, line 36: `dining` and `is_open_dining` are defined as `string` but they expect the digit `1`. would `true` and `false` be more user-friendly?

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
floppyears, line 89: data may be a list of a single object. When the user calls /locations - data holds a list and when the user calls /locations/123abc - data holds a single object.

src/test/groovy/edu/oregonstate/mist/locations/frontend/LocationDAOTest.groovy
floppyears, line 22: Yes

src/main/groovy/edu/oregonstate/mist/locations/frontend/db/LocationDAO.groovy
floppyears, line 49: The Resource class checks that the parameter is not null. This class won't be called with a null id. If for some reason the id is null, the "?." null safe operator takes care of that. 

This method actually makes an http request . This is the short notation. In groovy, you can convert a string to a url object. Calling "text" makes an http request and returns the string response.

swagger.yaml
floppyears, line 36: Good catch. I forgot to update the parameters section. Some of these features will be implemented in the next version of the API.

src/main/groovy/edu/oregonstate/mist/locations/frontend/jsonapi/ResourceObject.groovy
floppyears, line 6: Thanks for asking this question! I refreshed on def. Def is an alias of Object, but it has the benefit that allows us to use some of the dynamic typing. Since attributes can hold a variety of things depending on the API, I'm going to leave it as is.

ghost, line 6: you should be able to remove the comment in line 6 then

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
floppyears, line 138: Good job on asking this question. I did verify with groovysh when I was writing the code. Math.ceil( 5 / 2) returns 3.0 . Groovy does the rounding off when the result of an integer division is stored in an integer.

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
floppyears, line 177: I'm going to leave this as is. Here's why:
- Groovy actually expands findAll, but it uses metaprogramming and handles nulls
- The current performance of this API is 7ms , which is super fast. 
- I'm torn on converting this groovy code into loops because it adds more code to maintain, and if we end up using katharsis , we won't be needing this code.

ghost, line 177: i was thinking in terms of operations, not milliseconds. functional programming implementations do have some performance costs. i agree that it is unlikely to be significant however.

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
ghost, line 89: understood.

src/main/groovy/edu/oregonstate/mist/locations/frontend/db/LocationDAO.groovy
ghost, line 49: didn't know that. sorta like jquery. it might be worth adding a comment to make this clear to future maintainers.

PR #9, opened 2016-09-13
Feature/co 524 mockfor unit tests

tsoliangwu0130:


tsoliangwu0130:
I am not sure should we do the unit test for private methods or not? For example, like [getESSearchQuery](https://github.com/osu-mist/locations-frontend-api/blob/feature/CO-524-mockfor-unit-tests/src/main/groovy/edu/oregonstate/mist/locations/frontend/db/LocationDAO.groovy#L106). From what I understood, the private methods/fields should be hidden from the outside, or we should use **static** keyword?

PR #10, opened 2016-09-15
Integration tests

jaredkosanovic:


src/test/integration/configuration.json
```diff
@@ -0,0 +1,12 @@
+{
```
floppyears: I would suggest renaming this file as configuration-example.json or something similar. Then adding configuration.json to .gitignore. This will help prevent devs in the future from committing this file and credentials by mistake.

src/test/integration/checkResponseCode.py
```diff
@@ -0,0 +1,65 @@
...
+access_token_response = getAccessToken(access_token_url, config_data["client_id"], config_data["client_secret"]);
+
+access_token  = access_token_response["access_token"]
+locations_url = base_url + config_data["locations_endpoint"]
+params        = {"q": "Oxford"}
+
+response_code = getResponse(locations_url, access_token, params)
+print "response: ", response_code
+
+checkResponseCode(response_code)
```
floppyears: Since this code will mostly be test, I would vote for leveraging a test framework for the language being used. For example, https://docs.python.org/2/library/unittest.html . What I like about using a testing framework is that we can easily define startup / setup and teardown methods that can take care of initialization of objects / deps. The teardown methods help make sure we clean up resources / objects at the end. We also get an easy report at the end knowing which tests failed or succeeded.

src/test/integration/checkResponseCode.py
```diff
@@ -0,0 +1,65 @@
+import json
+import pycurl
+import urllib
+import StringIO
+
```
floppyears: Let's add a readme so that a new dev knows what command to use to run the tests. Another thing to keep in mind is whether some tests can be run in parallel. Maybe don't add support, but mention them as todo. This is something we could do to speed up integration tests in the future.

src/test/integration/checkResponseCode.py
```diff
@@ -0,0 +1,65 @@
...
+    curl.setopt(pycurl.WRITEFUNCTION, storage.write)
+
+    curl.perform()
+    response_code = curl.getinfo(pycurl.HTTP_CODE)
+    curl.close()
+
+    response = storage.getvalue()
+    return response_code
+
+def checkResponseCode(response_code):
```
floppyears: Hmm, your code does bring up a good point that I don't have an answer for yet. Should our integration tests go through the API gateway or not?

src/test/integration/checkResponseCode.py
```diff
@@ -0,0 +1,65 @@
...
+    assert response_code == 200, "Request unsuccessful"
```
floppyears: I like that we are checking the status code and using assertions. We should try and verify the body of the response as well as edge / error cases.

src/test/integration/README.md
```diff
@@ -0,0 +1,13 @@
+# Locations Frontend API Integration Tests
+
+This directory contains files that run integration tests against the locations frontend API. To run the tests, use Python. These libraries are required to run the integration tests:
```
floppyears, line 3: what version of python: 2.5, 2.7 or 3.x?

src/test/integration/configuration_load.py
```diff
@@ -0,0 +1,20 @@
+import json
+import requests
+
+def get_config_data():
+	config_data_file = open('configuration.json')
```
floppyears: Could we specify the configuration file as a parameter? See the example: https://github.com/osu-mist/dublabs-vcf-files/blob/develop/buildings_vcf_generator.py#L57 . This allows us to have the configuration files loaded externally in a path outside of the workspace.

src/test/integration/api_request.py
```diff
@@ -0,0 +1,53 @@
...
+def unauth_request():
+    query_params = {'q': 'Oxford'}
+    request = requests.get(get_url(), params=query_params)
+    return request.status_code
+    
+def not_found_request():
+    query_params = {'campus': 'Pluto'}
+    headers = {'Authorization': get_access_token()}
+    request = requests.get(get_url(), params=query_params, headers=headers)
+    return request.status_code
```
floppyears: For some of these requests that return error objects, could you assert that the fields that we expect are present?

src/test/integration/api_request.py
```diff
@@ -0,0 +1,53 @@
...
+import urllib2
+import ssl
+from configuration_load import *
+
+
+def good_request():
+    query_params = {'q': 'Oxford'}
+    headers = {'Authorization': get_access_token()}
+    request = requests.get(get_url(), params=query_params, headers=headers)
+    return request.status_code
```
floppyears, line 12: How about verifying that the links attribute is present in queries that return more than 10 results (along with links.self, links.next, links.prev)?

jaredkosanovic: The links attribute should be present all the time. The only difference with a query with more than 10 results is that "next" will not be null. I am adding tests for links assertions.

PR #11, opened 2016-10-07
Added more integration tests

jaredkosanovic:
Added integration tests that were present as unit tests in LocationDAOtest.groovy. Also added some similar tests that were created for terms-api


src/test/integration/integrationtests.py
```diff
@@ -6,9 +6,37 @@
 
 class gateway_tests(unittest.TestCase):
 
-	# Tests that a good request returns a 200
-	def test_success(self):
-		self.assertEqual(good_request(url, access_token), 200)
+	# Tests a single resource ID in different case styles
+	def test_id(self):
```
floppyears: Can you update these python files so that they are using spaces: https://www.python.org/dev/peps/pep-0008/#tabs-or-spaces ? Also can you set the default in your editor / IDE to be spaces?

jaredkosanovic: That's odd. I thought I had set Sublime to do spaces, and it did for api_request.py but not the other files. I'll update those.

PR #15, opened 2016-12-08
Feature/co 647 query with geo point

SuzyWu2014:


.gitignore
```diff
@@ -12,14 +15,19 @@ configuration.yaml
 *.ipr
 *.iws
 .idea/
+<<<<<<< HEAD
```
floppyears: This is a merge conflict

.gitignore
```diff
@@ -12,14 +15,19 @@ configuration.yaml
...
+out/
+>>>>>>> skeleton/master
 
 # Mac OS X
 .DS_Store
 
 # Build Properties
 build.properties
 
+<<<<<<< HEAD
```
floppyears: More merge conflicts

src/main/groovy/edu/oregonstate/mist/locations/frontend/db/LocationDAO.groovy
```diff
@@ -1,13 +1,15 @@
...
 import com.fasterxml.jackson.databind.ObjectMapper
+import org.joda.time.DateTime
 
 /**
  * Handles HTTP requests against ElasticSearch. Operation supported are:
  * search and findById.
  */
 class LocationDAO {
     private final Map<String, String> locationConfiguration
+    private final String DEFAULT_SEARCH_DISTANCE = "0.5km"
```
floppyears: I think this may be more useful as a value in the configuration.yaml file

src/main/groovy/edu/oregonstate/mist/locations/frontend/mapper/LocationMapper.groovy
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * Get attributes from JsonNode
+     * @param attr
+     * @param sort
+     * @return
+     */
+    private static Attributes getAttributes(JsonNode attr, JsonNode sort) {
+
+        Attributes attributes = new Attributes(
+                name: attr.get("name").asText(null),
```
floppyears: I think having a helper private method that you could pass in the field name and returns the output of attr.get($field).asText(null) would make this a bit DRYer.

src/main/groovy/edu/oregonstate/mist/locations/frontend/mapper/LocationMapper.groovy
```diff
@@ -0,0 +1,114 @@
...
+        attributes
+    }
+
+    /**
+     * Convert [JsonNode] to [String]
+     * @param node
+     * @return
+     */
+    private static List<String> getListOfString(JsonNode node) {
+
```
floppyears: empty line not needed?

src/main/groovy/edu/oregonstate/mist/locations/frontend/mapper/LocationMapper.groovy
```diff
@@ -0,0 +1,114 @@
...
+        List<String> result = null
+
+        if ( node != null ) {
```
floppyears: This can be simplified with:
List<String> result = node?.asList().collect { it.asText(null) }

src/main/groovy/edu/oregonstate/mist/locations/frontend/mapper/LocationMapper.groovy
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * Get OpenHours from JsonNode
+     * @param node
+     * @return
+     */
+    private static Map<Integer, List<DayOpenHours>> getOpenHours (JsonNode node) {
+        Map<Integer, List<DayOpenHours>> openHours = new HashMap<Integer, List<DayOpenHours>>()
+        if ( node != null ) {
+            (1..7).each {
+                openHours.put(it, getHoursList(node.get(Integer.toString(it))))
```
floppyears: At the end of this line, you are four levels deep in method calls. How about either using a helper function or a variable so that the code is a bit easier to read?

swagger.yaml
```diff
@@ -21,10 +21,48 @@ paths:
       parameters:
         - in: query
           name: q
-          description: "Query to search through buildings. Looks through the building name, abbreviation, summary and description"
+          description: "Query to search through buildings. Looks through the building name, abbreviation. Disabled when lat/lon parameter are presented."
```
magical: Why can't `q` be used with `lat/lon`?

src/main/groovy/edu/oregonstate/mist/locations/frontend/core/DayOpenHours.groovy
```diff
@@ -0,0 +1,25 @@
...
+     *
+     * An event can be overwritten by a new event by using the same UID with a
+     * different sequence
+     *
+     */
+    @JsonIgnore
+    String uid
+
+    @JsonIgnore
+    Integer sequence
```
magical: Are the uid and sequence fields necessary here? They shouldn't be present in the elasticsearch data.

src/main/groovy/edu/oregonstate/mist/locations/frontend/LocationsFrontEndApplication.groovy
```diff
@@ -37,15 +35,15 @@ class LocationsFrontEndApplication extends Application<LocationsFrontendConfigur
      */
     @Override
     public void run(LocationsFrontendConfiguration configuration, Environment environment) {
```
magical: It looks like you updated the skeleton.
You should add these lines from the skeleton here:
```
BuildInfoManager buildInfoManager = new BuildInfoManager()
environment.lifecycle().manage(buildInfoManager)
environment.jersey().register(new NotFoundExceptionMapper())
environment.jersey().register(new GenericExceptionMapper())
environment.jersey().register(new LoggingExceptionMapper<WebApplicationException>(){})
environment.jersey().register(new PrettyPrintResponseFilter())
environment.jersey().register(new InfoResource(buildInfoManager.getInfo()))
```

src/test/groovy/edu/oregonstate/mist/locations/frontend/MockUriInfo.groovy
```diff
@@ -1,50 +1,106 @@
 package edu.oregonstate.mist.locations.frontend
+
 import javax.ws.rs.core.UriInfo
 import javax.ws.rs.core.UriBuilder
 import javax.ws.rs.core.PathSegment
 import javax.ws.rs.core.MultivaluedMap
 import javax.ws.rs.core.MultivaluedHashMap
 
 class MockUriInfo implements UriInfo {
```
magical, line 9: We have two copies of this class now. One in src/test/groovy/edu/oregonstate/mist/api, and one in src/test/groovy/edu/oregonstate/mist/locations/frontend.

swagger.yaml
```diff
@@ -21,10 +21,48 @@ paths:
...
+            Kilometer - km or kilometers;
+            Meter - m or meters;
+            Centimeter - cm or centimeters;
+            Millimeter - mm or millimeters;
+            Nautical mile - NM, nmi or nauticalmiles
+          "
+          required: false
+          type: string
+        - in: query
+          name: isopen
```
magical: Why is distanceUnit in camelCase, but isopen all lowercase?

swagger.yaml
```diff
@@ -21,10 +21,48 @@ paths:
...
+          description: "True if only search for open restaurants/buildings. Default: 'false' to search for all the results"
```
magical: Drop the quotes around 'false'

swagger.yaml
```diff
@@ -179,6 +216,10 @@ definitions:
         type: array
         items:
           $ref: "#/definitions/DayHours"
+      distance:
+        description: "Distance from users' location in kilometer"
```
magical: Who is the user? This is only returned if lat/lon was specified, right?

src/test/groovy/edu/oregonstate/mist/locations/frontend/MockUriInfo.groovy
floppyears: @magical good catch. I think we can let this go for now. One of the things that I have mentioned to Jared in the past is that I'm not 100% sold with the locations-api being an API. At first it made sense, but the role of that codebase is just to query backends and output data that can be given to ElasticSearch. I think the locations backend lends itself better to be a something else than API. The logic behind the backend doesn't have to be executed via HTTP.

.gitignore
```diff
@@ -18,6 +18,10 @@ configuration.json
 .idea/
 classes/
 out/
+<<<<<<< HEAD
```
floppyears: These merge conflict artifacts are back. Let's remove them.

PR #16, opened 2017-01-19
CO-656 Add Dockerfile for intergration test

tsoliangwu0130:


src/test/integration/README.md
```diff
@@ -20,4 +20,25 @@ Successfully passing all the tests with the command above would output this resu
...
+
+```shell
+$ docker build -t location-frontend-tests .
+# Run the integration tests in Unix
+$ docker run -v "$PWD"/configuration.json:/usr/src/app/configuration.json:ro location-frontend-tests
+# Run the integration tests in Windows
+$ docker run -v c:\path\to\configuration.json:/c:\usr\src\app\configuration.json:ro location-frontend-tests
+```
+
+Currently, there is one failure test which will return wrong error code for POST method:
```
jaredkosanovic: Since we are expecting that test to fail, we should skip it until the root of the issue is fixed. Could you skip the test_verbs test and display a message like this?
'Apigee cache bug - JIRA ticket CO-699'

PR #20, opened 2017-04-11
Feature/co 745 hours

floppyears:


swagger.yaml
```diff
@@ -70,7 +70,7 @@ paths:
           type: string
         - in: query
```
jaredkosanovic, line 71: What is your collectionFormat for this query parameter? We originally talked about doing "multi" (ex. ?type=dining&type=building) but I wasn't sure if that's what you're implementing.

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
```diff
@@ -73,35 +59,25 @@ class LocationResource extends Resource {
...
     }
 
-    @Context
-    UriInfo uriInfo
-
     @GET
-    @Produces(MediaType.APPLICATION_JSON)
     @Timed
     Response list(@QueryParam('q') String q,
                   @QueryParam('campus') String campus, @QueryParam('type') String type,
```
jaredkosanovic, line 65: What if I want to query multiple types? I think we talked about doing something like type=blah&type=building which would be handled as an array of strings. Could you also update the swagger spec to indicate a collectionFormat for the type parameter?

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
```diff
@@ -3,33 +3,31 @@ package edu.oregonstate.mist.locations.frontend.resources
...
 
 @Path("locations")
 @Produces(MediaType.APPLICATION_JSON)
+@PermitAll
 class LocationResource extends Resource {
     private static final Logger LOGGER = LoggerFactory.getLogger(LocationResource.class)
 
     public static final ArrayList<String> ALLOWED_CAMPUSES = ["corvallis", "extension"]
-    public static final ArrayList<String> ALLOWED_TYPES = ["building", "dining"]
+    public static final ArrayList<String> ALLOWED_TYPES = ["building", "dining", "cultural-centers"]
```
jaredkosanovic: building and dining is singular, while cultural-centers is plural. Should we just make it "cultural-center" to be consistent?

swagger.yaml
floppyears: I thought about this during the implementation. I went with just allowing a single type to make the implementation a bit simpler. For now, if needed the consumer can make multiple API calls. Down the road once we migrate to using an ES library, if needed we can support multiple type parameters.

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
floppyears: Sure

swagger.yaml
```diff
@@ -133,6 +133,115 @@ paths:
...
+          description: "Page number of results. Used to paginate through results."
+          required: false
+          type: string
+        - $ref: '#/parameters/pretty'
+      responses:
+        "200":
+          description: "Successful response"
+          schema:
+            $ref: "#/definitions/ServicesResultObjects"
+        "404":
```
jaredkosanovic: Is this applicable here? If a /services query returned nothing, it will still return a 200 with an empty data array.

swagger.yaml
```diff
@@ -165,6 +274,13 @@ definitions:
...
           $ref: "#/definitions/ResourceObject"
 
+  ServicesResultObject:
+    properties:
+      links:
+        $ref: "#/definitions/Links"
+      data:
+        $ref: "#/definitions/ServicesResourceObject"
+
   ResourceObject:
```
jaredkosanovic: You have ResourceObject/ResultObject and ServicesResourceObject/ServicesResultObject. Could you qualify the locations resource and result object? Ex LocationsResultObject.

configuration-example.yaml
```diff
@@ -47,43 +48,10 @@ locations:
   esUrl: http://elastic-search-url.example.com
   esIndex: elastic-search-index
   estype: elastic-search-type
+  esIndexService: elastic-search-services-index
+  estypeService: elastic-search-services-type
+  gatewayUrl: https://api.example.com/v1/
```
jaredkosanovic: I left a comment in the configs repo about this. I was wondering why you're adding back gatewayUrl since we already have endpointUri. I think we changed the code a few months ago to remove gatewayUrl dependencies.

swagger.yaml
```diff
@@ -173,24 +289,42 @@ definitions:
...
     properties:
-      name:
+      links:
+        $ref: "#/definitions/Links"
+      data:
+        type: array
+        items:
+          $ref: "#/definitions/ServicesResourceObject"
+
+  ServicesResourceObject:
```
jaredkosanovic, line 298: Should the servicesresourceobject contain a relationships object?

swagger.yaml
```diff
@@ -275,18 +448,13 @@ definitions:
...
-        description: Build time as milliseconds since the unix epoch.
-      commit:
+
+  ServiceLinks:
+    properties:
+      self:
         type: string
-        description: Git commit hash of the revision which was built.
-      documentation:
+      location:
```
jaredkosanovic: I don't think we should have this in the links section. I would recommend having the location link in the relationship section similar to how location objects link to services. Also, json api spec only allows links objects to contain "self" and "related" links. You could have the api respond with something like:
```
"relationships": {
    "location": {
        "links":{
            "related": "api.o.e/v1/locations/idofparentlocation"
        }
    }
}

swagger.yaml
floppyears: Hmm, the jsonapi.org spec wasn't specific as to what was the preferred way to map back to the locations object. I went with a links.location . It seemed cleaner and simple. We could go with the relationships object, I just don't think it makes sense to have both.

jaredkosanovic: You could go with links.related and link that to the locations object. I agree jsonapi isn't too clear though.

jaredkosanovic (approved):

PR #23, opened 2017-04-24
add geometries

jaredkosanovic:


swagger.yaml
```diff
@@ -322,9 +322,27 @@ definitions:
...
-        type: string
+        type: number
+        format: float
       longitude:
-        type: string
+        type: number
+        format: float
+      geometry:
+        properties:
+          rings:
```
floppyears: How about if we try using the format & type specified by geoJson for the geometry object: https://tools.ietf.org/html/rfc7946#section-3.1.5 ?

floppyears (approved):

PR #28, opened 2017-06-13
Feature/co 815 gender inclusive rr implementation

jaredkosanovic:


src/main/groovy/edu/oregonstate/mist/locations/frontend/db/LocationDAO.groovy
```diff
@@ -66,7 +66,7 @@ class LocationDAO {
         // generate ES query to search for locations
         def esQuery = getESSearchQuery(q, null, null,
                                        null, null, null,
-                                       isOpen, pageNumber, pageSize)
+                                       isOpen, null, pageNumber, pageSize)
```
floppyears, line 69: Hmm, I wonder if we should change the getESSearchQuery method to use a map with named arguments? It would simplify the code that calls that method: http://docs.groovy-lang.org/latest/html/documentation/#_named_arguments . Depending on when we plan to refactor the code to use the ES library / sdk, we could hold off on the code change.

floppyears (approved):

PR #29, opened 2017-06-30
allow other type, check pageSize

jaredkosanovic:


src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
```diff
@@ -70,6 +72,10 @@ class LocationResource extends Resource {
                   @QueryParam('giRestroom') Boolean giRestroom) {
 
         try {
+            if (pageSize > MAX_PAGE_SIZE) {
```
floppyears: Is this something that we should have as part of the skeleton? The various APIs could then overwrite the variable or use a configuration value to provide another value.

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
```diff
@@ -26,8 +26,10 @@ import java.util.regex.Pattern
 class LocationResource extends Resource {
     private static final Logger LOGGER = LoggerFactory.getLogger(LocationResource.class)
 
+    private final Integer MAX_PAGE_SIZE = 10000
     public static final ArrayList<String> ALLOWED_CAMPUSES = ["corvallis", "extension"]
-    public static final ArrayList<String> ALLOWED_TYPES = ["building", "dining", "cultural-center"]
+    public static final ArrayList<String> ALLOWED_TYPES = ["building", "dining",
```
floppyears: Should the swagger spec, readme or smartdoc be updated to mention the allowed types or max_page size?

floppyears (approved):

PR #37, opened 2017-10-17
CO-927: use Elasticsearch Java API

magical:


jaredkosanovic:
I realized that putting a page[size] of 0 causes a divide-by-zero error in [this line](https://github.com/osu-mist/locations-frontend-api/blob/feature/CO-927-elasticsearch-java-api/src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy#L198). I'm thinking we should add some extra logic at the [skeleton level](https://github.com/osu-mist/locations-frontend-api/blob/develop/src/main/groovy/edu/oregonstate/mist/api/Resource.groovy) to make sure the page[size] is greater than 0 and less than the [max page size](https://github.com/osu-mist/locations-frontend-api/blob/develop/src/main/groovy/edu/oregonstate/mist/api/Resource.groovy#L28).

magical:
@jaredkosanovic Looks like we already check if page[size] is negative https://github.com/osu-mist/web-api-skeleton/blame/master/src/main/groovy/edu/oregonstate/mist/api/Resource.groovy#L176. Should probably change it to exclude 0 as well.

jaredkosanovic (approved):

PR #41, opened 2017-11-03
CO-948: Add integration test for isOpen

tsoliangwu0130:


src/test/integration/integrationtests.py
```diff
@@ -1,8 +1,19 @@
...
+                        check_ssl, \
+                        get_buildings_with_services, \
+                        id_request, \
+                        not_found_request, \
+                        query_request, \
+                        response_time, \
+                        results_with_links, \
+                        unauth_request
+from configuration_load import get_access_token, get_single_resource_id, get_url
+
```
jaredkosanovic, line 16: extra blank line?

src/test/integration/integrationtests.py
```diff
@@ -221,11 +262,12 @@ def test_ssl_v3(self):
             self.skipTest('SSLv3 support not available')
         self.assertFalse(check_ssl(ssl.PROTOCOL_SSLv3, locations_url, access_token))
 
+
```
jaredkosanovic, line 256: extra blank line?

src/test/integration/integrationtests.py
```diff
@@ -145,14 +156,44 @@ def test_parking(self):
...
+                open_hours = open_resource['attributes']['openHours']
+                self.assertIsNotNone(open_hours)
+
+                # If there are not only one open hours, current time should be in one of these open hours intervals
+                if len(open_hours[weekday]) > 1:
+                    is_open = False
+                    for open_hour in open_hours[weekday]:
+                        if open_hour['start'][:-1] < now < open_hour['end'][:-1]:
+                            is_open = True
+                            self.assertTrue(is_open)
```
jaredkosanovic: What if none of the time slots contain the current time? This test would never fail, it just wouldn't run, yes? If none of the time slots surround the current time, that should be a fail.

src/test/integration/integrationtests.py
```diff
@@ -162,6 +163,37 @@ def test_parking(self):
...
+            for open_resource in all_open_resources['data']:
+                # Test that only open resources are returned when they should be and each open resource has a related open hours
+                open_hours = open_resource['attributes']['openHours']
+                self.assertIsNotNone(open_hours)
+
+                # If there are not only one open hours, current time should be in one of these open hours intervals
+                if len(open_hours[weekday]) > 1:
+                    is_open = False
+                    for open_hour in open_hours[weekday]:
+                        if open_hour['start'][:-1] < now < open_hour['end'][:-1]:
```
magical: is the `[:-1]` to strip the trailing Z? can we do that explicitly instead? `open_hour['start'].rstrip('Z')`. or just add a trailing Z to `now`.

src/test/integration/integrationtests.py
magical: yes. i think this line needs to be dedented by a couple blocks. it seems like what we really want to say is `self.assertTrue( any(open_hour['start'] <= now+'Z' <= open_hour['end'] for open_hour in open_hours[weekday]) )`

src/test/integration/integrationtests.py
tsoliangwu0130: hmm, in pep8/flake8, we generally use two blank lines between two def sections. See error code E305 here: https://github.com/PyCQA/pycodestyle/blob/master/docs/intro.rst#error-codes

src/test/integration/integrationtests.py
tsoliangwu0130: @jaredkosanovic good catch. Will fix it.

src/test/integration/integrationtests.py
tsoliangwu0130: Agree. Will fix.

src/test/integration/integrationtests.py
```diff
@@ -183,13 +183,9 @@ def test_resource(resource_url):
 
                 # If there are not only one open hours, current time should be in one of these open hours intervals
                 if len(open_hours[weekday]) > 1:
```
magical: I think this if statement is unnecessary now.

tsoliangwu0130: ha you are right!

jaredkosanovic (approved):

PR #42, opened 2017-11-21
Expose isOpen field dynamically

tsoliangwu0130:


src/main/groovy/edu/oregonstate/mist/locations/frontend/mapper/LocationMapper.groovy
```diff
@@ -50,5 +52,28 @@ class LocationMapper {
 
         // hashCode is used as metadata in ES. No need to expose it
         ro?.attributes?.remove("hashCode")
+
+        // expose isOpen field
+        def now = DateTime.now(DateTimeZone.UTC)
```
magical, line 57: this function would be easier to test if `now` were supplied as an argument

magical: nevermind. thanks for pointing out the existence of the DateTimeUtils.setCurrentMillisFixed function :)

tsoliangwu0130:
Not merging this PR for now because of the current caching proxy would affect it. We should merge this PR until we implement some cache-control functionality.

PR #44, opened 2018-01-11
CO-1000 Implement functionality to provide GeoJSON data

tsoliangwu0130:
Instead of having new endpoint for `/geojson`, finally decide to use `geojson` as a parameter.


src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
```diff
@@ -275,6 +292,68 @@ class LocationResource extends Resource {
 
     }
 
+    private static toGeoJson(ResultObject resultObject) {
```
jaredkosanovic: Could you add a comment for each of these new geojson methods?

jaredkosanovic (approved):

readme.md
```diff
@@ -336,3 +336,341 @@ This resource returns the information for a given building:
   }
 }
 ```
+
+## GeoJSON
+
+Locations API now provides [GeoJSON](https://tools.ietf.org/html/rfc7946) format for `/locations` and `/locations/{id}` endpoints with the parameter `geojson=true`.
```
magical: Drop "now" - it's going to get out of date very quickly.

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
```diff
@@ -275,6 +292,80 @@ class LocationResource extends Resource {
...
+     */
+    private static toGeoJson(ResultObject resultObject) {
+        def geojsonResultObject
+        def ro = resultObject?.data
+
+        if (ro instanceof List) {
+            geojsonResultObject = new GeoFeatureCollection(type: "FeatureCollection")
+            ro.each {
+                geojsonResultObject?.features += adjustGeoFeature(it)
+            }
```
magical: This could be written more succinctly as `geojsonResultObject.features = ro.collect { adjustGeoFeature(it) }`

Also, the `?.` isn't necessary because you just constructed the object, so it can't be null.

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
```diff
@@ -275,6 +292,80 @@ class LocationResource extends Resource {
...
+        }
+
+        if (geoPolygon && geoPoint) {
+            geometry = new Geometries()
+            geometry?.type = "GeometryCollection"
+            geometry?.geometries = [geoPolygon, geoPoint]
+        } else if (!(geoPolygon || geoPoint)) {
+            geometry = null
+        } else {
+            def coordinates = [geoPolygon, geoPoint].find { it }
```
magical: Ha, clever. You could also use the `?:` operator here: `def coordinates = geoPolygon ?: geoPoint`

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
```diff
@@ -275,6 +292,80 @@ class LocationResource extends Resource {
...
+                ]
+            )
+        }
+
+        if (geoPolygon && geoPoint) {
+            geometry = new Geometries()
+            geometry?.type = "GeometryCollection"
+            geometry?.geometries = [geoPolygon, geoPoint]
+        } else if (!(geoPolygon || geoPoint)) {
+            geometry = null
```
magical: This case doesn't seem necessary. I think it would be more clear to remove this `else if` and turn the `else` below into `else if (geoPolygon || geoPoint)`

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
```diff
@@ -275,6 +292,80 @@ class LocationResource extends Resource {
...
+     *
+     * @param ro
+     * @return
+     */
+    private static GeoFeature adjustGeoFeature(ResourceObject ro) {
+        def geojsonResultObject = new GeoFeature(type: "Feature")
+
+        def geoPolygon = null
+        def geoPoint = null
+        def geometry
```
magical: Isn't `def geometry` equivalent to `def geometry = null`? Why omit `null` here but not in the previous two declarations?

src/main/groovy/edu/oregonstate/mist/api/jsonapi/GeoJsonResultObject.groovy
```diff
@@ -0,0 +1,36 @@
+package edu.oregonstate.mist.api.jsonapi
```
magical: Why are these classes in the jsonapi package?

src/main/groovy/edu/oregonstate/mist/locations/frontend/resources/LocationResource.groovy
```diff
@@ -275,6 +292,80 @@ class LocationResource extends Resource {
...
+
+        // adjust geoPolygon
+        if (ro?.attributes?.geometry?.type && ro?.attributes?.geometry?.coordinates) {
```
magical: It's great that you're coding defensively by using `?.` everywhere (something I often forget to do) but if `ro` is null we're just going to end up returning an object full of nulls. I'd almost rather crash in that case.

magical (commented):
Looks good. Nice documentation :+1: 

I left a bunch of stylistic comments in LocationResource. Feel free to disregard those, if you want.

PR #55, opened 2018-04-24
CO-1076 add query params for parking space

tsoliangwu0130:


swagger.yaml
```diff
@@ -88,6 +88,24 @@ paths:
           required: false
           collectionFormat: multi
           type: string
+        - in: query
```
jaredkosanovic, line 91: Could you make the names and descriptions of these new query parameters more detailed? For example, what are your thoughts about making them called the same as the fields that relate to? (adaParkingSpaceCount, etc.)

src/main/groovy/edu/oregonstate/mist/locations/frontend/db/LocationDAO.groovy
```diff
@@ -254,6 +256,18 @@ class LocationDAO {
             query.must(parkingZoneGroupQuery)
         }
 
+        if (ada) {
+            query.must(QueryBuilders.rangeQuery("attributes.adaParkingSpaceCount").gt(ada))
```
jaredkosanovic: What do you think making this a greater than or equal to instead? If I want parking lots with 1 or more ada parking spaces (or moto or ev, etc.) then I would need to put in an `ada` parameter of 0, but that returns every location (ones with `null` for the ada space count too).

jaredkosanovic (changes requested):

src/main/groovy/edu/oregonstate/mist/locations/frontend/db/LocationDAO.groovy
tsoliangwu0130: Good catch. Will fix.

jaredkosanovic (approved):

PR #60, opened 2018-06-22
Feature/co 1012

gmtcrary:
CO-1012

I refactored the mapping logic so all of the JSON handling fits in one screen.
I also added a docopt at the top so the command line arguments are easier to understand for a first time user.
Added a requirements.txt for the docopt import.


DataDiscrepancyUtil/dataDiffCheck.py
```diff
@@ -1,61 +1,60 @@
...
-            old_bdata_name_dict[old_building['attributes']
-                                ['name']] = old_building['id']
+    
+    if(args['<old_data_path>'] == args['<new_data_path>']):
+        print "These are the same file..."
+        sys.exit(1)
+    else:
+
+        with open(args['<old_data_path>'], "r") as old_json_file:
+            with open(args['<new_data_path>'], "r") as new_json_file:
```
tsoliangwu0130: You could open multiple files like this:

```python
with open('a', 'w') as a, open('b', 'w') as b:
    do_something()
```

https://stackoverflow.com/questions/4617034/how-can-i-open-multiple-files-using-with-open-in-python

DataDiscrepancyUtil/dataDiffCheck.py
```diff
@@ -1,61 +1,60 @@
...
+        new_data_path: File path of new building json
+"""
+# old.json being the data from https://api.oregonstate.edu/v1/locations?page[size]=10000&type=building
+# new.json being the data from https://localhost:8082/api/v0/locations?page[size]=10000&type=building - locations-frontend-api commit b1bec1e013cf29f16a0a01b9dd7c3777e3b4e192
+
+from docopt import docopt
+print(docopt(__doc__, version='1.0.0rc2'))
+
+args = docopt(__doc__, version='1.0.0rc2')
+
```
tsoliangwu0130, line 17: According to PEP8 standard, module level import should be at top of the file. See error [E402](http://pep8.readthedocs.io/en/release-1.7.x/intro.html#error-codes).

DataDiscrepancyUtil/dataDiffCheck.py
```diff
@@ -80,14 +79,13 @@
                 if rekeyCheckDict.has_key(bname):
```
tsoliangwu0130: `has_key` has been deprecated, instead, we should use `in` here.

tsoliangwu0130 (changes requested):
The code looks good, there are only a few [PEP8](https://www.python.org/dev/peps/pep-0008/) warnings. Could you fix these warning?

gmtcrary:
Fixed, @tsoliangwu0130.

DataDiscrepancyUtil/dataDiffCheck.py
gmtcrary: Thanks for showing me this!

tsoliangwu0130 (approved):
Looks good!

PR #78, opened 2020-04-16
CO-1690: V2 openapi design

jemisonf:
I didn't feel like we needed an extensive overhaul on this API doc, so I worked pretty directly off the original swagger.yaml file without a ton of structural changes. The biggest ones are:
* Adding included resources to /locations and /locations/{locationId}
* Updating query parameters to be more consistent with our style guide
* Updating to use openapi 3 syntax
* Trying to add some quality of life improvements to the documentation where possible

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -390,7 +390,7 @@ components:
         attributes:
           $ref: '#/components/schemas/ServicesAttributes'
         links:
-          $ref: '#/components/schemas/ServicePaginationLinks'
+          $ref: '#/components/schemas/SelfLink'
     LocationsResourceObject:
       properties:
         id:
@@ -428,7 +428,7 @@ components:
         attributes:
           $ref: '#/components/schemas/ServicesAttributes'
         links:
-          $ref: '#/components/schemas/ServicePaginationLinks'
+          $ref: '#/components/schemas/SelfLink'
         relationships:
           $ref: '#/components/schemas/RelationshipsInServices'
     Meta:
@@ -461,14 +461,18 @@ components:
           type: string
           description: Name of the location.
           example: Dixon Recreation Center
-        bannerAbbreviation:
-          type: string
-          description: Abbreviation for the location
-          example: DxRC
-        arcGisAbbreviation:
-          type: string
-          description: Abbreviation for the location
-          example: DxRC
+        abbreviations:
+          type: object
+          description: possible abbreviations for the location
+          properties:
+            banner:
+              type: string
+              description: Banner abbreviation for the location
+              example: DxRC
+            arcGis:
+              type: string
+              description: ArcGIS abreviation for the location
+              example: DxRC
         latitude:
           type: number
           format: float
@@ -544,20 +548,24 @@ components:
           type: string
           example: '541-111-1111'
           description: a fax number that can be used to reach the location
-        thumbnailUris:
-          type: array
-          description: an array of URLs for thumbnail images of the location
-          items:
-            type: string
-            format: uri
-            example: https://example.oregonstate.edu/image.jpg
-        imageUris:
-          description: an array of URLs for images of the location
-          type: array
-          items:
-            type: string
-            format: uri
-            example: https://example.oregonstate.edu/image.jpg
+        images:
+          type: object
+          description: full size and thumbnail image URLs for the location
+          properties:
+            fullSize:
+              description: an array of URLs for images of the location
+              type: array
+              items:
+                type: string
+                format: uri
+                example: https://example.oregonstate.edu/image.jpg
+            thumbnails:
+              type: array
+              description: an array of URLs for thumbnail images of the location
+              items:
+                type: string
+                format: uri
+                example: https://example.oregonstate.edu/image.jpg
         departments:
           type: array
           description: departments related to the location
@@ -571,6 +579,7 @@ components:
           example: https://example.oregonstate.edu
         sqft:
           type: number
+          format: integer
           description: the size in square feet of the location
           example: 12345
           nullable: true
@@ -593,24 +602,27 @@ components:
           type: string
           enum: [building, dining, other, parking, cultural-center]
           description: the type of the building
-        giRestroomCount:
-          description: Number of gender inclusive restrooms available in the location.
-          type: number
-          format: integer
-        giRestroomLimit:
-          description: >-
-            If true, access to this location's restrooms may be limited.
-            A residence hall might restrict bathroom use or building access to
-            residents only.
-          type: boolean
-        giRestroomLocations:
-          description: >-
-            Room numbers of gender inclusive restrooms within location.
-            Some restrooms may not be ADA compliant.
-          type: array
-          items:
-            type: string
-            example: 0111A
+        giRestrooms:
+          type: object
+          properties:
+            count:
+              description: Number of gender inclusive restrooms available in the location.
+              type: number
+              format: integer
+            limit:
+              description: >-
+                If true, access to this location's restrooms may be limited.
+                A residence hall might restrict bathroom use or building access to
+                residents only.
+              type: boolean
+            locations:
+              description: >-
+                Room numbers of gender inclusive restrooms within location.
+                Some restrooms may not be ADA compliant.
+              type: array
+              items:
+                type: string
+                example: 0111A
         synonyms:
           type: array
           description: other terms that can be used for the location
@@ -650,18 +662,22 @@ components:
             - B3
             - C
             - R
-        evParkingSpaceCount:
-          description: Count of electric vehicle parking spaces if the location is parking
-          type: number
-          format: integer
-        adaParkingSpaceCount:
-          description: Count of ADA parking spaces if the location is parking
-          type: number
-          format: integer
-        motorcycleParkingSpaceCount:
-          description: Count of motorcycle parking spaces if the location is parking.
-          type: number
-          format: integer
+        parkingSpaces:
+          type: object
+          required: [evSpaceCount, adaSpaceCount, motorcycleSpaceCount]
+          properties:
+            evSpaceCount:
+              description: Count of electric vehicle parking spaces if the location is parking
+              type: number
+              format: integer
+            adaSpaceCount:
+              description: Count of ADA parking spaces if the location is parking
+              type: number
+              format: integer
+            motorcycleSpaceCount:
+              description: Count of motorcycle parking spaces if the location is parking.
+              type: number
+              format: integer
         weeklyMenu:
           description: Link to a weekly menu if the location is dining.
           type: string
@@ -724,8 +740,9 @@ components:
     DayHours:
       description: >-
         This array represents open hours for a location, relative to the date
-        that the API returns the response. The first element represents the current day,
-        the second represents the next day etc. All times are in UTC.
+        that the API returns the response. The array has seven elements: the 
+        first element represents the current day, the second represents the next 
+        day etc. All times are in UTC.
       type: array
       items:
         $ref: '#/components/schemas/TimeSlot'
@@ -757,10 +774,6 @@ components:
           format: uri
           nullable: true
           description: The next page of data
-    ServicePaginationLinks:
-      properties:
-        self:
-          type: string
     ErrorResult:
       properties:
         errors:
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -700,43 +700,37 @@ components:
     RelationshipsInLocations:
       properties:
         services:
-          $ref: '#/components/schemas/LocationsRelationship'
+          properties:
+            data:
+              type: array
+              items:
+                type: object
+                properties:
+                  id:
+                    type: string
+                    example: 5c4f1efcbaada0c63e
+                  type:
+                    type: string
+                    enum: [service]
+            links:
+              $ref: '#/components/schemas/SelfLink'
     RelationshipsInServices:
       properties:
-        locations:
-          $ref: '#/components/schemas/ServicesRelationship'
-    ServicesRelationship:
-      properties:
-        data:
-          type: array
-          items:
-            $ref: '#/components/schemas/LocationsResourceObjectIdentifier'
-        links:
-          $ref: '#/components/schemas/SelfLink'
-    LocationsRelationship:
-      properties:
-        data:
-          type: array
-          items:
-            $ref: '#/components/schemas/ServicesResourceObjectIdentifier'
-        links:
-          $ref: '#/components/schemas/SelfLink'
-    LocationsResourceObjectIdentifier:
-      properties:
-        id:
-          type: string
-          example: 5c4f1efcbaada0c63e
-        type:
-          type: string
-          enum: [location]
-    ServicesResourceObjectIdentifier:
-      properties:
-        id:
-          type: string
-          example: 5c4f1efcbaada0c63e
-        type:
-          type: string
-          enum: [service]
+        services:
+          properties:
+            data:
+              type: array
+              items:
+                type: object
+                properties:
+                  id:
+                    type: string
+                    example: 5c4f1efcbaada0c63e
+                  type:
+                    type: string
+                    enum: [location]
+            links:
+              $ref: '#/components/schemas/SelfLink'
     DayHours:
       description: >-
         This array represents open hours for a location, relative to the date
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -35,14 +35,14 @@ paths:
             type: string
         - in: query
           name: filter[latitude]
-          description: 'Latitude of the location. Should be used with lon parameter'
+          description: 'Latitude of the location. Should be used with `filter[longitude]` parameter'
           required: false
           schema:
             type: number
             format: float
         - in: query
           name: filter[longitude]
-          description: 'Longitude of the location. Should be used with lat parameter'
+          description: 'Longitude of the location. Should be used with `filter[latitude]` parameter'
           required: false
           schema:
             type: number
@@ -67,6 +67,8 @@ paths:
             Centimeter - `cm` or `centimeters`;
             Millimeter - `mm` or `millimeters`;
             Nautical mile - `NM`, `nmi` or `nauticalmiles`
+
+            Should be used with the `filter[distance]` parameter
           '
           required: false
           schema:
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -142,6 +142,7 @@ paths:
           schema:
             type: number
             format: integer
+            example: 1
         - in: query
           name: filter[motoParkingSpaceCount][gte]
           description: Only include locations with motorcycle parking space greater than or equal to this amount
@@ -149,6 +150,7 @@ paths:
           schema:
             type: number
             format: integer
+            example: 1
         - in: query
           name: filter[evParkingSpaceCount][gte]
           description: Only include locations with electric vehicle parking space greater than or equal to this amount
@@ -156,6 +158,7 @@ paths:
           schema:
             type: number
             format: integer
+            example: 1
         - in: query
           name: filter[bannerAbbreviation]
           description: Search by banner abbreviation
@@ -352,6 +355,12 @@ components:
           schema:
             $ref: '#/components/schemas/ErrorResult'
   schemas:
+    LocationType:
+      type: string
+      enum: [location]
+    ServiceType:
+      type: string
+      enum: [service]
     LocationResult:
       properties:
         links:
@@ -395,8 +404,7 @@ components:
         id:
           type: string
         type:
-          type: string
-          enum: [service]
+          $ref: '#/components/schemas/ServiceType'
         attributes:
           $ref: '#/components/schemas/ServicesAttributes'
         links:
@@ -406,10 +414,9 @@ components:
         id:
           type: string
           example: 5c4f1efcbaada0c63e
-          description: ID of a location object
+          description: ID of a service object
         type:
-          type: string
-          enum: [service]
+          $ref: '#/components/schemas/LocationType'
         attributes:
           $ref: '#/components/schemas/LocationsAttributes'
         links:
@@ -421,8 +428,7 @@ components:
           example: 5c4f1efcbaada0c63e
           description: ID of a location object
         type:
-          type: string
-          enum: [service]
+          $ref: '#/components/schemas/LocationType'
         attributes:
           $ref: '#/components/schemas/LocationsAttributes'
         links:
@@ -450,8 +456,7 @@ components:
           example: 5c4f1efcbaada0c63e
           description: ID of a service object
         type:
-          type: string
-          enum: [service]
+          $ref: '#/components/schemas/ServiceType'
         attributes:
           $ref: '#/components/schemas/ServicesAttributes'
         links:
@@ -772,8 +777,7 @@ components:
                     type: string
                     example: 5c4f1efcbaada0c63e
                   type:
-                    type: string
-                    enum: [service]
+                    $ref: '#/components/schemas/ServiceType'
             links:
               $ref: '#/components/schemas/SelfLink'
     RelationshipsInServices:
@@ -789,8 +793,7 @@ components:
                     type: string
                     example: 5c4f1efcbaada0c63e
                   type:
-                    type: string
-                    enum: [location]
+                    $ref: '#/components/schemas/LocationType'
             links:
               $ref: '#/components/schemas/SelfLink'
     DayHours:
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -803,6 +803,8 @@ components:
         first element represents the current day, the second represents the next 
         day etc. All times are in UTC.
       type: array
+      minItems: 7
+      maxItems: 7
       items:
         $ref: '#/components/schemas/TimeSlot'
     TimeSlot:
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -458,15 +458,19 @@ components:
         name:
           type: string
           description: Name of the location.
+          example: Dixon Recreation Center
         abbreviation:
           type: string
           description: Abbreviation for the location
+          example: DxRC
         latitude:
           type: number
           format: float
+          example: 44.563137
         longitude:
           type: number
           format: float
+          example: -123.278729
         geometry:
           description: >-
             an array of arrays of (latitude, longitude) coordinates 
@@ -491,6 +495,7 @@ components:
         zoneSummary:
           type: string
           description: the description of the parking zone of the group
+          example: 'Zone: Austin Hall'
         description:
           type: object
           description: miscellaneous information about the location in HTML and text format
@@ -498,9 +503,17 @@ components:
             html:
               type: string
               description: the description for the location, formatted using HTML tags
+              example: >-
+                <h4>\n\tAccessibility:</h4>
+                \n<p>\n\tENTRIES: North entry: ramp up to main floor.<br />
+                \n\tFLOORS: All floors; elevator (no Braille).</p>\n
             text:
               type: string
               description: the desciption for the location, formatted without HTML tags
+              example: >-
+                \r\n\tAccessibility:\r\n\r\n\r\n\t
+                ENTRIES: North entry: ramp up to main floor.\r\n\t
+                FLOORS: All floors; elevator (no Braille).\r\n
         address:
           type: string
           example: 12345 NE STREET RD
@@ -553,9 +566,11 @@ components:
         sqft:
           type: string
           description: the size in square feet of the location
+          example: '12345'
         calendar:
           type: string
           description: a link, if available, to a calendar file for the location
+          example: https://example.oregonstate.edu/cal.ics
         campus:
           type: string
           example: corvallis
@@ -649,6 +664,7 @@ components:
         name:
           type: string
           description: Name of the Service.
+          example: Dixon Pool
         tags:
           description: Tags associated with this location
           type: array
@@ -660,7 +676,7 @@ components:
     RelationshipsInLocations:
       properties:
         services:
-          $ref: '#/components/schemas/ServicesRelationship'
+          $ref: '#/components/schemas/LocationsRelationship'
     RelationshipsInServices:
       properties:
         locations:
@@ -670,15 +686,33 @@ components:
         data:
           type: array
           items:
-            $ref: '#/components/schemas/ResourceObjectIdentifier'
+            $ref: '#/components/schemas/LocationsResourceObjectIdentifier'
         links:
           $ref: '#/components/schemas/SelfLink'
-    ResourceObjectIdentifier:
+    LocationsRelationship:
+      properties:
+        data:
+          type: array
+          items:
+            $ref: '#/components/schemas/ServicesResourceObjectIdentifier'
+        links:
+          $ref: '#/components/schemas/SelfLink'
+    LocationsResourceObjectIdentifier:
       properties:
         id:
           type: string
+          example: 5c4f1efcbaada0c63e
         type:
           type: string
+          enum: [location]
+    ServicesResourceObjectIdentifier:
+      properties:
+        id:
+          type: string
+          example: 5c4f1efcbaada0c63e
+        type:
+          type: string
+          enum: [service]
     DayHours:
       description: >-
         This array represents open hours for a location, relative to the date
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -34,19 +34,13 @@ paths:
           schema:
             type: string
         - in: query
-          name: filter[latitude]
-          description: 'Latitude of the location. Should be used with `filter[longitude]` parameter'
+          name: filter[coordinates]
+          description: A (latitude,longitude) pair of coordinates
           required: false
           schema:
-            type: number
-            format: float
-        - in: query
-          name: filter[longitude]
-          description: 'Longitude of the location. Should be used with `filter[latitude]` parameter'
-          required: false
-          schema:
-            type: number
-            format: float
+            type: string
+            pattern: '^-?\d+\.\d+,-?\d+\.\d+$'
+            example: 44.567509,-123.286651
         - in: query
           name: filter[distance]
           description: 'The radius of the circle centred on the specified location. Default: 2'
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -479,6 +479,7 @@ components:
     LocationsAttributes:
       properties:
         name:
+          nullable: false
           type: string
           description: Name of the location.
           example: Dixon Recreation Center
@@ -489,23 +490,28 @@ components:
             banner:
               type: string
               description: Banner abbreviation for the location
+              nullable: true
               example: DxRC
             arcGis:
               type: string
               description: ArcGIS abreviation for the location
+              nullable: true
               example: DxRC
         latitude:
           type: number
           format: float
+          nullable: true
           example: 44.563137
         longitude:
           type: number
           format: float
+          nullable: true
           example: -123.278729
         geometry:
           description: >-
             an array of arrays of (latitude, longitude) coordinates 
             representing the shapes that make up the location
+          nullable: true
           properties:
             type:
               type: string
@@ -524,12 +530,14 @@ components:
                     type: number
                     format: float
         zoneSummary:
+          nullable: true
           type: string
           description: the description of the parking zone of the group
           example: 'Zone: Austin Hall'
         description:
           type: object
           description: miscellaneous information about the location in HTML and text format
+          nullable: true
           properties:
             html:
               type: string
@@ -548,22 +556,27 @@ components:
         address:
           type: string
           example: 12345 NE STREET RD
+          nullable: true
           description: the street address of the location
         city:
           type: string
           example: Corvallis
+          nullable: true
           description: the city where the location is located
         state:
           type: string
           example: OR
+          nullable: true
           description: the state where the location is located
         zip:
           type: string
           example: '97330'
+          nullable: true
           description: the zip code of the location
         telephone:
           type: string
           example: '541-111-1111'
+          nullable: true
           description: a phone number that can be used to reach the location
         fax:
           type: string
@@ -572,10 +585,12 @@ components:
         images:
           type: object
           description: full size and thumbnail image URLs for the location
+          nullable: false
           properties:
             fullSize:
               description: an array of URLs for images of the location
               type: array
+              nullable: false
               items:
                 type: string
                 format: uri
@@ -583,6 +598,7 @@ components:
             thumbnails:
               type: array
               description: an array of URLs for thumbnail images of the location
+              nullable: false
               items:
                 type: string
                 format: uri
@@ -590,12 +606,14 @@ components:
         departments:
           type: array
           description: departments related to the location
+          nullable: false
           items:
             type: string
             example: College of Engineering
         website:
           type: string
           description: the website of the location
+          nullable: true
           format: uri
           example: https://example.oregonstate.edu
         sqft:
@@ -607,12 +625,14 @@ components:
         calendar:
           type: string
           description: a link, if available, to a calendar file for the location
+          nullable: true
           format: uri
           example: https://example.oregonstate.edu/cal.ics
         campus:
           type: string
           example: corvallis
           description: the OSU campus where the location is located
+          nullable: false
           enum:
             - corvallis
             - extension
@@ -622,30 +642,36 @@ components:
         type:
           type: string
           enum: [building, dining, other, parking, cultural-center]
+          nullable: false
           description: the type of the building
         giRestrooms:
           type: object
+          nullable: false
           properties:
             count:
               description: Number of gender inclusive restrooms available in the location.
               type: number
+              nullable: true
               format: integer
             limit:
               description: >-
                 If true, access to this location's restrooms may be limited.
                 A residence hall might restrict bathroom use or building access to
                 residents only.
+              nullable: true
               type: boolean
             locations:
               description: >-
                 Room numbers of gender inclusive restrooms within location.
                 Some restrooms may not be ADA compliant.
               type: array
+              nullable: true
               items:
                 type: string
                 example: 0111A
         synonyms:
           type: array
+          nullable: false
           description: other terms that can be used for the location
           items:
             type: string
@@ -658,10 +684,12 @@ components:
             This is only returned when the query parameters lat and lon are specified.
             The default distance unit is mile
           type: number
+          nullable: true
           format: float
           example: 1.8
         tags:
           description: Tags associated with this location
+          nullable: false
           type: array
           items:
             type: string
@@ -674,6 +702,7 @@ components:
         parkingZoneGroup:
           description: Zonegroup of location if location is parking
           type: string
+          nullable: true
           enum:
             - A1
             - A2
@@ -691,18 +720,22 @@ components:
               description: Count of electric vehicle parking spaces if the location is parking
               type: number
               format: integer
+              nullable: true
             adaSpaceCount:
               description: Count of ADA parking spaces if the location is parking
               type: number
               format: integer
+              nullable: true
             motorcycleSpaceCount:
               description: Count of motorcycle parking spaces if the location is parking.
               type: number
               format: integer
+              nullable: true
         weeklyMenu:
           description: Link to a weekly menu if the location is dining.
           type: string
           format: uri
+          nullable: true
           example: https://app.uhds.oregonstate.edu/api/dining/weeklymenu/index?loc=1
     ServicesAttributes:
       properties:
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -165,21 +165,14 @@ paths:
             example: DxRC
         - in: query
           name: filter[arcGisAbbreviation]
-          description: Search by ArcGIS abbreviat ion
+          description: Search by ArcGIS abbreviation
           required: false
           schema:
             type: string
             example: DxRC
         - $ref: '#/components/parameters/pageNumber'
         - $ref: '#/components/parameters/pageSize'
-        - in: query
-          name: include
-          required: false
-          schema:
-            type: array
-            items:
-              type: string
-              enum: [services]
+        - $ref: '#/components/parameters/locationsIncluded'
       responses:
         '200':
           description: Successful response
@@ -199,6 +192,7 @@ paths:
       operationId: getLocationById
       parameters:
         - $ref: '#/components/parameters/locationId'
+        - $ref: '#/components/parameters/locationsIncluded'
       responses:
         '200':
           description: Successful response
@@ -285,6 +279,18 @@ paths:
           $ref: '#/components/responses/500'
 components:
   parameters:
+    locationsIncluded:
+      in: query
+      name: include
+      required: false
+      description: >-
+        Add "services" to have the response include all service
+        resources for this location
+      schema:
+        type: array
+        items:
+          type: string
+          enum: [services]
     locationId:
       name: locationId
       in: path
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -34,6 +34,12 @@ paths:
           schema:
             type: string
         - in: query
+          name: filter[name]
+          description: Search buildings by name
+          required: false
+          schema:
+            type: string
+        - in: query
           name: filter[coordinates]
           description: A (latitude,longitude) pair of coordinates
           required: false
@@ -53,42 +59,21 @@ paths:
           name: filter[distanceUnit]
           description: |
             Distance can be specified in following units:
-            Mile - `mi` or `miles` (default);
-            Yard - `yd` or `yards`;
-            Feet - `ft` or `feet`;
-            Inch - `in` or `inch`;
-            Kilometer - `km` or `kilometers`;
-            Meter - `m` or `meters`;
-            Centimeter - `cm` or `centimeters`;
-            Millimeter - `mm` or `millimeters`;
-            Nautical mile - `NM`, `nmi` or `nauticalmiles`
+            - `mi` - miles
+            - `yd` - yards
+            - `ft` - feet
+            - `in` - inches
+            - `km` - kilometers
+            - `m` - meters
+            - `cm` - centimeters
+            - `mm` - millimeters
+            - `nmi` - nautical miles
 
             Should be used with the `filter[distance]` parameter
           required: false
           schema:
             type: string
-            enum:
-              [
-                mi,
-                miles,
-                yd,
-                yards,
-                ft,
-                feet,
-                in,
-                inch,
-                km,
-                kilometers,
-                m,
-                meters,
-                cm,
-                centimeters,
-                mm,
-                millimeters,
-                nmi,
-                NM,
-                nauticalmiles,
-              ]
+            enum: [mi, yd, ft, in, km, m, cm, mm, nmi]
         - $ref: '#/components/parameters/isOpen'
         - in: query
           name: filter[campus][oneOf]
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -355,6 +355,21 @@ components:
           schema:
             $ref: '#/components/schemas/ErrorResult'
   schemas:
+    CoordinatePair:
+      description: a pair of latitude, longitude coordinates
+      type: object
+      required: [lat, long]
+      properties:
+        lat:
+          example: 44.562934
+          type: number
+          format: float
+          description: the latitude value of the coordinate pair
+        long:
+          example: -123.281901
+          type: number
+          format: float
+          description: the longitude value of the coordinate pair
     LocationType:
       type: string
       enum: [location]
@@ -508,16 +523,8 @@ components:
               description: ArcGIS abreviation for the location
               nullable: true
               example: DxRC
-        latitude:
-          type: number
-          format: float
-          nullable: true
-          example: 44.563137
-        longitude:
-          type: number
-          format: float
-          nullable: true
-          example: -123.278729
+        coordinates:
+          $ref: '#/components/schemas/CoordinatePair'
         geometry:
           description: >-
             an array of arrays of (latitude, longitude) coordinates 
@@ -536,10 +543,7 @@ components:
                 description: Array of points that make up a ring.
                 type: array
                 items:
-                  type: array
-                  items:
-                    type: number
-                    format: float
+                  $ref: '#/components/schemas/CoordinatePair'
         zoneSummary:
           nullable: true
           type: string
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -185,14 +185,14 @@ paths:
           $ref: '#/components/responses/400'
         '500':
           $ref: '#/components/responses/500'
-  /locations/{locationID}:
+  /locations/{locationId}:
     get:
       tags:
         - locations
       description: 'Finds location by ID'
       operationId: getLocationByID
       parameters:
-        - name: locationID
+        - name: locationId
           in: path
           description: 'ID of location to be searched by'
           required: true
@@ -209,7 +209,7 @@ paths:
           $ref: '#/components/responses/404'
         '500':
           $ref: '#/components/responses/500'
-  /locations/{locationID}/services:
+  /locations/{locationId}/services:
     get:
       tags:
         - locations
@@ -217,7 +217,7 @@ paths:
       description: 'Lists services related to a location'
       operationId: getServices
       parameters:
-        - name: locationID
+        - name: locationId
           in: path
           description: 'ID of location'
           required: true
@@ -267,14 +267,14 @@ paths:
           $ref: "#/components/responses/400"
         '500':
           $ref: "#/components/responses/500"
-  /services/{serviceID}:
+  /services/{serviceId}:
     get:
       tags:
         - services
       description: 'Finds service by ID'
       operationId: getServiceById
       parameters:
-        - name: serviceID
+        - name: serviceId
           in: path
           description: 'ID of location to be searched by'
           required: true
@@ -628,27 +628,34 @@ components:
       properties:
         self:
           type: string
-    Error:
-      type: object
+    ErrorResult:
+      properties:
+        errors:
+          type: array
+          items:
+            $ref: '#/components/schemas/ErrorObject'
+    ErrorObject:
       properties:
         status:
-          type: integer
-          format: int32
-          description: HTTP status code.
-        developerMessage:
           type: string
-          description: An error string aimed at developers.
-        userMesage:
+          description: HTTP status code
+          example: '123'
+        title:
           type: string
-          description: An error string aimed at end users.
+          description: A short, user readable summary of the error
+          example: Not Found
         code:
-          type: integer
-          description: Error code.
-        details:
           type: string
-          description: A link to further information about the error.
-    Errors:
-      title: ArrayOfErrors
-      type: array
-      items:
-        $ref: '#/components/schemas/Error'
+          description: An application-specific error code
+          example: '1234'
+        detail:
+          type: string
+          description: A long description of the error that may contain instance-specific
+            details
+        links:
+          properties:
+            about:
+              type: string
+              format: uri
+              description: A link to further information about the error
+              example: https://developer.oregonstate.edu/documentation/error-reference#1234
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -35,7 +35,13 @@ paths:
             type: string
         - in: query
           name: filter[name]
-          description: Search buildings by name
+          description: Search for a building with this exact name
+          required: false
+          schema:
+            type: string
+        - in: query
+          name: filter[name][fuzzy]
+          description: Search for a building with a name similiar to the query
           required: false
           schema:
             type: string
@@ -476,10 +482,19 @@ components:
                   items:
                     type: number
                     format: float
-        summary:
+        zoneSummary:
           type: string
+          description: the description of the parking zone of the group
         description:
-          type: string
+          type: object
+          description: miscellaneous information about the location in HTML and text format
+          properties:
+            html:
+              type: string
+              description: the description for the location, formatted using HTML tags
+            text:
+              type: string
+              description: the desciption for the location, formatted without HTML tags
         address:
           type: string
           example: 12345 NE STREET RD
@@ -534,6 +549,7 @@ components:
           description: the size in square feet of the location
         calendar:
           type: string
+          description: a link, if available, to a calendar file for the location
         campus:
           type: string
           example: corvallis
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -458,38 +458,54 @@ components:
           type: string
         address:
           type: string
+          example: 12345 NE STREET RD
         city:
           type: string
+          example: Corvallis
         state:
           type: string
+          example: OR
         zip:
           type: string
+          example: '97330'
         telephone:
           type: string
+          example: 541-111-1111
         fax:
           type: string
+          example: 541-111-1111
         thumbnails:
           type: array
           items:
             type: string
+            format: url
+            example: https://example.oregonstate.edu/image.jpg
         images:
           type: array
           items:
             type: string
+            format: url
+            example: https://example.oregonstate.edu/image.jpg
         departments:
           type: array
           items:
             type: string
+            example: College of Engineering
         website:
           type: string
+          format: url
+          example: https://example.oregonstate.edu
         sqft:
           type: string
         calendar:
           type: string
         campus:
           type: string
+          example: Corvallis
         type:
           type: string
+          enum: [building, dining, other, parking, cultural-center]
+          example: building
         giRestroomCount:
           description: Number of gender inclusive restrooms available in the location.
           type: integer
@@ -504,10 +520,12 @@ components:
             Room numbers of gender inclusive restrooms within location.
             Some restrooms may not be ADA compliant.
           type: string
+          example: 0111A, 0101, 0102
         synonyms:
           type: array
           items:
             type: string
+            example: "Gymnasium"
         openHours:
           $ref: '#/components/schemas/DayHours'
         distance:
@@ -516,14 +534,17 @@ components:
                 The default distance unit is mile '
           type: number
           format: float
+          example: 1.8
         tags:
           description: 'Tags associated with this location'
           type: array
           items:
             type: string
+            example: cultural-center
         parkingZoneGroup:
           description: 'Zonegroup of location if location is parking'
           type: string
+          example: A1
         evParkingSpaceCount:
           description: 'Count of electric vehicle parking spaces if the location is parking'
           type: integer
@@ -536,6 +557,8 @@ components:
         weeklyMenu:
           description: 'Link to a weekly menu if the location is dining.'
           type: string
+          format: url
+          example: https://app.uhds.oregonstate.edu/api/dining/weeklymenu/index?loc=1
     ServicesAttributes:
       properties:
         name:
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -570,9 +570,10 @@ components:
           format: uri
           example: https://example.oregonstate.edu
         sqft:
-          type: string
+          type: number
           description: the size in square feet of the location
-          example: '12345'
+          example: 12345
+          nullable: true
         calendar:
           type: string
           description: a link, if available, to a calendar file for the location
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -742,10 +742,12 @@ components:
         name:
           type: string
           description: Name of the Service.
+          nullable: false
           example: Dixon Pool
         tags:
           description: Tags associated with this location
           type: array
+          nullable: false
           items:
             type: string
             example: recreation
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1,12 +1,10 @@
-swagger: '2.0'
+openapi: '3.0.2'
 info:
-  version: "0.0.1"
+  version: "2.0.0"
   title: Locations API
   description: Get information about OSU locations.
-host: 'api.oregonstate.edu'
-schemes:
-  - https
-basePath: /v2
+servers: 
+  - url: https://api.oregonstate.edu/v2
 security:
   - OAuth2:
     - full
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -1,279 +1,662 @@
-swagger: '2.0'
+openapi: '3.0.2'
```
jemisonf: using 3.0.2 instead of 3.0.3 because my vscode extensions don't support 3.0.3. If that's an issue I can work around the extensions problem, but I don't think the two are very different.

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+          name: filter[parkingZoneGroup][oneOf]
+          description: >
+            Locations in results should match the parkingZoneGroup
+            in this query parameter.
+          required: false
+          schema:
+            type: array
+            items:
+              type: string
+              # what are the possible values for this?
```
jemisonf: I wasn't sure what the options for this were, and figured it would be better to just ask someone who has access to the source data instead of trying to guess based on online OSU resources

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+              enum: [services]
       responses:
-        '200':
-          description: Successful response
+        "200":
+          description: "Successful response"
+          content:
+            "application/json":
+              schema:
+                $ref: "#/components/schemas/LocationsResultObjects"
```
jemisonf: Lots of stuff like this in the new openapi 3 spec that kinda just feels like boilerplate

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+    get:
+      tags:
+       - services
+      summary: Get services
+      description: "Gets all `Service` objects. Can be filtered by parameters."
+      operationId: getServicesList
+      parameters:
+        - in: query
+          name: filter[query]
+          description: "Query to search through buildings. Looks through the building name, abbreviation."
```
jemisonf: I don't think this description is right (ie it's searching services, not buildings) but I'm not sure what the correct distinction is

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+        type: boolean
+  securitySchemes:
+    OAuth2:
+      type: oauth2
+      flows: 
+        clientCredentials:
+          tokenUrl: https://api.oregonstate.edu/oauth2/token
+          refreshUrl: https://api.oregonstate.edu/oauth2/token
+          scopes: 
+            full: full access to the API
```
jemisonf: Not positive this is the correct way to implement this, but it's my best try based on the [available documentation](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#oauth-flows-object)

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+        links:
+          $ref: "#/components/schemas/Links"
+        data:
+          type: array
+          items:
+            $ref: "#/components/schemas/LocationsResourceObject"
+        included:
+          type: array
+          items:
+            $ref: "#/components/schemas/ServicesIncluded"
```
jemisonf: if the performance of getting a compound document is bad, one option that I think would work is only allowing compound documents with `/locations/{locationId}`, assuming that's something people would find useful

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+      summary: Get locations
+      description: "Gets all `Location` objects. Can be filtered by parameters."
+      operationId: getLocations
       parameters:
-        - $ref: '#/parameters/pageNumber'
-        - $ref: '#/parameters/pageSize'
         - in: query
-          name: species
-          description: Search by pet species
+          name: filter[query]
```
kalcicd: For this and the `/services` query param, maybe we could rename them as `filter[broadSearch]` or just `filter[search]` to avoid any confusion between the 'query' query parameter and all other query parameters

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+          description: "Query to search through buildings based on building name and abbreviation"
```
kalcicd: this could just read `Search through buildings based on building name and abbreviation`

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+            items:
+              type: string
+              enum:
+                - corvallis
+                - extension
+                - cascades
+                - hmsc
+                - other
+        - in: query
+          name: filter[type]
```
kalcicd, line 102: this param is missing a `required:` field

openapi.yaml
tsoliangwu0130: I think we should use 3.0.0 for now since our new skeleton is implemeted base on this version. If we wanna upgrade openapi version, we should upgrade skeleton first and then pull the change from there.

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
-host: api.oregonstate.edu
-basePath: /v1
-externalDocs:
-  description: GitHub Repository
-  url: https://github.com/osu-mist/express-api-skeleton
-produces:
-  - application/json
-consumes:
-  - application/json
+  version: "2.0.0"
```
tsoliangwu0130: This should be `v2`. Also, looks like there are many fields (e.g. `license`, `contact`, etc.) missing from our skeleton. I recommend merging this file with the [openapi from the skeleton (openapi3 branch)](https://github.com/osu-mist/express-api-skeleton/blob/openapi3/openapi.yaml) so that you won't need to add them all manually. Ideally, you should use the skeleton openapi as template and modify if needed.

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+  title: Locations API
+  description: Get information about OSU locations.
```
tsoliangwu0130: And also services.

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+  /locations:
     get:
-      summary: List all pets
       tags:
-        - pets
-      description: Get all pets
-      operationId: getPets
+       - locations
+      summary: Get locations
+      description: "Gets all `Location` objects. Can be filtered by parameters."
```
tsoliangwu0130: Double-quotes are not necessary. Take a look at style guide: https://github.com/osu-mist/style-guides/blob/master/OpenAPI.md#quoting-conventions

openapi.yaml
tsoliangwu0130: I'd try avoiding this kind of ambiguous parameter in our newer version APIs. I think you can split them to `filter[name]` and `filter[location]`. (I don't think we should use `buildingName` since not every location is a building.)

You remind me another issue regarding abbreviation. We have been told that Banner and ARCGIS sometimes use different abbreviations for buildings. We should separate them to two different fields in resource and add filter for each of them.

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+                $ref: "#/components/schemas/Errors"
+        "500":
+          description: "Internal Server Error"
+          content:
+            "application/json":
+              schema:
+                $ref: "#/components/schemas/Errors"
+components:
+  parameters:
+    pretty:
```
kalcicd: I haven't seen this parameter in any newer api that returns a JSON object. This is more of a question for @tsoliangwu0130 , but do we have users who need this feature? Otherwise I think we could remove it.

openapi.yaml
kalcicd: Looks good to me

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
-          description: Search by pet species
+          name: filter[query]
+          description: "Query to search through buildings based on building name and abbreviation"
           required: false
-          type: string
+          schema:
+            type: string
+        - in: query
+          name: filter[latitude]
+          description: "Latitude of the location. Should be used with lon parameter"
```
tsoliangwu0130: Should be used with `filter[longitude]` parameter.

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+          required: false
+          schema:
+            type: number
+            format: float
+        - in: query
+          name: filter[longitude]
+          description: "Longitude of the location. Should be used with lat parameter"
```
tsoliangwu0130: Should be used with `filter[latitude]` parameter.

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+          required: false
+          schema:
+            type: number
+            format: float
+        - in: query
+          name: filter[distance]
+          description: "The radius of the circle centred on the specified location. Default: 2"
```
tsoliangwu0130: You probably wanna mention this should be used with `filter[distanceUnit]`.

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+          required: false
+          schema:
+            type: number
+            format: float
```
tsoliangwu0130, line 66: Can you add default value for this param?

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+          description: "Distance can be specified in following units:
+            Mile - mi or miles (default);
+            Yard - yd or yards;
+            Feet - ft or feet;
+            Inch - in or inch;
+            Kilometer - km or kilometers;
+            Meter - m or meters;
+            Centimeter - cm or centimeters;
+            Millimeter - mm or millimeters;
+            Nautical mile - NM, nmi or nauticalmiles
```
tsoliangwu0130: Can you fix the style here by following our style guide? https://github.com/osu-mist/style-guides/blob/master/OpenAPI.md#quoting-conventions

> Descriptions for abbreviated words, typically enum values, should use backticks (`) around the abbreviation.

openapi.yaml
kalcicd: By compound documents, are you referring to the plural ResultObjects that references the resource object? I'm not sure if using $ref in the swagger doc will affect the speed of our responses

kalcicd (changes requested):

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+              - kilometers
+              - m
+              - meters
+              - cm
+              - centimeters
+              - mm
+              - millimeters
+              - nmi
+              - NM
+              - nauticalmiles
```
tsoliangwu0130: You can put them in a normal array since you have listed them in description. Example: https://github.com/osu-mist/persons-api/blob/v2/openapi.yaml#L118-L142

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+        - in: query
+          name: filter[isOpen]
+          description: "Set to true to only search for open restaurants/buildings. Default: false to search for all the results"
+          required: false
+          schema:
+            type: boolean
+        - in: query
+          name: filter[campus][oneOf]
+          description: "Name of the campus. Valid options are: corvallis, extension, cascades, hmsc, and other."
```
tsoliangwu0130: No need to list valid option here since you have had `enum` list below.

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+            items:
+              type: string
+              enum:
+                - building
+                - dining
+                - cultural-center
+                - parking
+                - other
+        - in: query
+          name: filter[giRestroom]
```
tsoliangwu0130: Just my two cents, I think `filter[hasGiRestroom]` is easier to understand.

openapi.yaml
tsoliangwu0130: Screen Shot 2020-04-16 at 4 38 02 PM

tsoliangwu0130: I sent you and @kalcicd and link via Slack for the source data. Let me know if you can access it or now.

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+              type: string
+              # what are the possible values for this?
+        - in: query
+          name: filter[adaParkingSpaceCount][gte]
+          description: "Only include locations with ADA parking space greater than or equal to this amount"
+          required: false
+          schema:
+            type: integer
+        - in: query
+          name: filter[motorcycleParkingSpaceCount][gte]
```
tsoliangwu0130: Not sure, personally I think it's too long 😛 How about `filter[motoParkingSpaceCount][gte]`?

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+          schema:
+            type: integer
+        - in: query
+          name: filter[evParkingSpaceCount][gte]
+          description: "Only include locations with electric vehicle parking space greater than or equal to this amount"
+          required: false
+          schema:
+            type: integer
+        - in: query
+          name: filter[abbreviation]
```
tsoliangwu0130: See my comment above.

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+          required: false
+          schema:
+            type: string
+        - in: query
+          name: page[size]
+          description: "Number of results to return. Used in pagination."
+          required: false
+          schema:
+            type: integer
+            maximum: 10000
```
tsoliangwu0130: I don't think it's a good idea to return 10,000 results in a response. Let's use the same default/max/min pagination value with other APIs: https://github.com/osu-mist/terms-api/blob/752ea46da0045124953a6d08df2ff5b32e16a517/openapi.yaml#L143-L159

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+            default: 10
+        - in: query
+          name: page[number]
+          description: "Page number of results. Used to paginate through results."
+          required: false
+          schema:
+            type: integer
+        - $ref: '#/components/parameters/pretty'
```
tsoliangwu0130: What's this for?

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+        - in: query
+          name: include
+          required: false
+          schema:
+            type: array
+            default: []
```
tsoliangwu0130: Hmm, it's not required but has default value? I guess it's a bit confusing to me.

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+            items:
+              type: string
+              enum: [services]
       responses:
-        '200':
-          description: Successful response
+        "200":
```
tsoliangwu0130: Should be single quotes: https://github.com/osu-mist/style-guides/blob/master/OpenAPI.md#quoting-conventions

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+    OAuth2:
+      type: oauth2
+      flows: 
+        clientCredentials:
+          tokenUrl: https://api.oregonstate.edu/oauth2/token
+          refreshUrl: https://api.oregonstate.edu/oauth2/token
+          scopes: 
+            full: full access to the API
+  schemas:
+    LocationsResultObject:
```
tsoliangwu0130: Can you rename these by following the style guide: https://github.com/osu-mist/style-guides/blob/master/OpenAPI.md#result-object--resource-object

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
       responses:
-        '200':
-          description: Successful response
+        "200":
+          description: "Successful response"
+          content:
+            "application/json":
+              schema:
+                $ref: "#/components/schemas/LocationsResultObjects"
+        "404":
```
tsoliangwu0130: In what circumstance this endpoint will return a 404? Also, I belive we need 400 for this endpoint.

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+            "application/json":
+              schema:
+                $ref: "#/components/schemas/Error"
+        "500":
+          description: "Internal Server Error"
+          content:
+            "application/json":
+              schema:
+                $ref: "#/components/schemas/Errors"
+  /locations/{locationID}:
```
tsoliangwu0130: `locationID` -> `locationId`

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
-      operationId: postPet
+            type: string
+      responses:
+        "200":
+          description: "Successful response"
+          content:
+            "application/json":
+              schema:
+                $ref: "#/components/schemas/LocationsResultObject"
+        "500":
```
tsoliangwu0130: We also need a 404 for this endpoint.

tsoliangwu0130 (changes requested):
I figure I should pause my review here for now or the PR will be a monster.

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+            type: string
+        - $ref: '#/components/parameters/pretty'
+      responses:
+        "200":
+          description: "Successful response"
+          content:
+            "application/json":
+              schema:
+                $ref: "#/components/schemas/ServicesResultObjects"
+        "500":
```
tsoliangwu0130: You need a 400 for this endpoint as well.

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
-          description: Conflict
+            type: boolean
+        - in: query
+          name: page[size]
+          description: "Number of results to return. Used in pagination. Default value is 10."
+          required: false
+          schema:
+            type: string
+        - in: query
+          name: page[number]
```
tsoliangwu0130: Some parameters (or all?) are reused in many places, you should reuse them like here: https://github.com/osu-mist/express-api-skeleton/blob/2e334d2061be607ddb09e417cbdfe4d5febcc81a/openapi.yaml#L30-L31

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+            "application/json":
+              schema:
+                $ref: "#/components/schemas/ServicesResultObjects"
+        "500":
+          description: "Internal Server Error"
+          content:
+            "application/json":
+              schema:
+                $ref: "#/components/schemas/Errors"
+  /services/{serviceID}:
```
tsoliangwu0130: `serviceID` -> `serviceId`

openapi.yaml
tsoliangwu0130: @kalcicd good point. I also left my comment regarding this in the above comment. Long story short, this parameter was used in the very first version of our old DW skeleton and JSON response was not prettified by default. However, since we have switched to the new express skeleton, I think this param is not necessary anymore.

openapi.yaml
tsoliangwu0130: You can just use what we have defined in the skeleton: https://github.com/osu-mist/express-api-skeleton/blob/2e334d2061be607ddb09e417cbdfe4d5febcc81a/openapi.yaml#L121-L127

openapi.yaml
```diff
@@ -1,279 +1,662 @@
...
+        links:
+          $ref: "#/components/schemas/ServiceLinks"
+        relationships:
+          $ref: "#/components/schemas/RelationshipsInServices"
+    SelfLink:
+      properties:
+        self:
+          type: string
+          description: Url to this resource / location.
+    Attributes:
```
tsoliangwu0130: Can you add example, description, and format/pattern (if applied) to all of these attributes?

tsoliangwu0130 (changes requested):

openapi.yaml
jemisonf: Thanks, I didn't realize we already had an openapi v3 doc for the skeleton

openapi.yaml
jemisonf: Compound documents meaning a response object from the API that includes all of the resources referenced in the `relationships` keys of the main resource(s) being fetched. [JSON:API docs here](https://jsonapi.org/format/#document-compound-documents). The concern would be that in the API itself we'd need to fetch a bunch of extra resources that we wouldn't have to if we weren't using a compound document.

openapi.yaml
jemisonf: I like `filter[search]`. I think you're right that `query` isn't really specific enough.

jemisonf: I do actually like the idea of having a single search field that can look across multiple attributes though. If I wanted to search for "Dixon" it's intuitive to me that I'd send `filter[search]=dixon` (using David's suggested name) whereas with `name` and `location` I wouldn't be super clear what the best on the use is. Might just be me though. Thoughts @tsoliangwu0130?

I also wonder if it would make sense to add those in addition to the `filter[search]` field so users can look for an exact or fuzzy match (depending how we implement it) for a specific field if that's what they want to do.

openapi.yaml
jemisonf: If we have to use both at the same time, do you know if there's a good way to encode that in the spec? I'll investigate to see if I can find something.

openapi.yaml
jemisonf: Yeah, I would agree

openapi.yaml
jemisonf: That makes sense, I think `moto` is a little more ambiguous than `motorcycle`, but the description makes it clear what it's referring to

openapi.yaml
jemisonf: This one seems like an OK length. I feel like if we shortened it more it would be inconsistent with the other filters.

openapi.yaml
jemisonf: That sounds good. We don't even have 10,000 of any of the resources in this API, right?

openapi.yaml
jemisonf: I wanted to make sure it was clear that if you didn't send it it wouldn't include anything by default. Probably fine to take it out.

openapi.yaml
jemisonf: I moved the pagination parameters and the isOpen parameters to the parameters definition section. I don't think there are others that are reused.

openapi.yaml
kalcicd: What if we consolidated both parameters into something like `filter[coordinates]` and have it accept an array containing the latitude and longitude?

tsoliangwu0130: @kalcicd it's a very good idea. We can then use `pattern` to limit the format. I like it 👍

jemisonf: I like that, I'll add it in. Format will be `/\d\.\d,\d\.\d/`

tsoliangwu0130: I think it should be `^\d+.\d+,\d+.\d+$`.

openapi.yaml
```diff
@@ -1,279 +1,684 @@
...
+      properties:
+        services:
+            $ref: '#/components/schemas/ServicesRelationship'
+    RelationshipsInServices:
+      properties:
+        locations:
+            $ref: '#/components/schemas/ServicesRelationship'
+    ServicesRelationship:
+      properties:
+        data:
```
kalcicd: I'd consider adding a links object here to display self links for the related resource

openapi.yaml
jemisonf: Oh yeah true. Do you know if this is right?
```
        - in: query
          name: filter[coordinates]
          description: A (latitude, longitude) pair of coordinates
          required: false
          schema:
            type: string
            pattern: ^\d\.\d, \d\.\d$
```

it's not showing any errors but the pattern doesn't show up in swagger UI from what I can tell

tsoliangwu0130: You need to quote it: https://github.com/osu-mist/terms-api/blob/master/openapi.yaml#L43

openapi.yaml
```diff
@@ -1,279 +1,684 @@
...
+          schema:
+            type: number
+            format: float
+        - in: query
+          name: filter[longitude]
+          description: 'Longitude of the location. Should be used with `filter[latitude]` parameter'
+          required: false
+          schema:
+            type: number
+            format: float
```
kalcicd: As Liang mentioned, longitude and latitude must be used together. The only way that comes to mind where we could enforce this in this document is if we grouped these two parameters into one `filter[coordinates]` param which takes an array of the latitude and longitude.

kalcicd: oops, just saw that you already addressed this

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
-host: api.oregonstate.edu
-basePath: /v1
 externalDocs:
   description: GitHub Repository
-  url: https://github.com/osu-mist/express-api-skeleton
-produces:
-  - application/json
-consumes:
-  - application/json
+  url: https://github.com/osu-mist/locations-frontend-api
```
tsoliangwu0130, line 15: My bad, I think we should move this file to [locations-api](https://github.com/osu-mist/locations-api) since this API should be renamed to Locations API to avoid confusion. We can do the openapi review here, once it's done, we can then close this PR and open a new one there.

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+      description: Gets all `Location` objects. Can be filtered by parameters.
+      operationId: getLocations
       parameters:
-        - $ref: '#/parameters/pageNumber'
-        - $ref: '#/parameters/pageSize'
         - in: query
-          name: species
-          description: Search by pet species
+          name: filter[search]
+          description: 'Search through buildings based on building name and abbreviation'
```
tsoliangwu0130: You don't need quotes for this string. Take a look at here: https://github.com/osu-mist/style-guides/blob/master/OpenAPI.md#quoting-conventions

Can you also fix other places among this file?

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+            type: string
+            enum: [mi, miles,
+                   yd, yards,
+                   ft, feet,
+                   in, inch,
+                   km, kilometers,
+                   m, meters,
+                   cm, centimeters,
+                   mm, millimeters,
+                   nmi, NM, nauticalmiles]
```
tsoliangwu0130: I'm not sure why would we need 2-3 aliases for each unit. I think one for each should be enough. What do you think?

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+            Mile - `mi` or `miles` (default);
+            Yard - `yd` or `yards`;
+            Feet - `ft` or `feet`;
+            Inch - `in` or `inch`;
+            Kilometer - `km` or `kilometers`;
+            Meter - `m` or `meters`;
+            Centimeter - `cm` or `centimeters`;
+            Millimeter - `mm` or `millimeters`;
+            Nautical mile - `NM`, `nmi` or `nauticalmiles`
+
```
tsoliangwu0130, line 80: Based on our [style guide](https://github.com/osu-mist/style-guides/blob/master/OpenAPI.md#quoting-conventions), it should be something like:

```yaml
  description: |
    Some sentence about this parameter.
    - `Code1` - description
    - `Code2` - description
```

By what you defined here, the available codes will be `Mile`, `Yard`, etc, which is not correct.

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+              enum:
+                - corvallis
+                - extension
+                - cascades
+                - hmsc
+                - other
+        - in: query
+          name: filter[type]
+          required: false
+          description: 'Type of location. Valid options are: building, dining, cultural-center, parking, and other.'
```
tsoliangwu0130: You have an enum below so I think you don't need to specify valid options here again.

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+              type: string
+              enum:
+                - building
+                - dining
+                - cultural-center
+                - parking
+                - other
+        - in: query
+          name: filter[hasGiRestroom]
+          description: >
```
tsoliangwu0130: You should use `>-` here. Take a look at the difference [here](https://github.com/osu-mist/style-guides/blob/master/OpenAPI.md#line-length--multiline-strings). Make sure you fix other places as well.

openapi.yaml
tsoliangwu0130: As mentioned, we will need two different filters for abbreviations since they are different between Banner and arcGIS. Let call `bannerAbbr` and `arcGisAbbr` or something similar.

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+          $ref: "#/components/responses/404"
+        '500':
+          $ref: "#/components/responses/500"
+  /services:
+    get:
+      tags:
+       - services
+      summary: Get services
+      description: 'Gets all `Service` objects. Can be filtered by parameters.'
+      operationId: getServicesList
```
tsoliangwu0130: getServicesList -> getServices.

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
-      summary: Add a new pet
-      operationId: postPet
+          $ref: '#/components/responses/500'
+  /locations/{locationId}:
+    get:
       tags:
-        - pets
+        - locations
+      description: 'Finds location by ID'
+      operationId: getLocationByID
```
tsoliangwu0130: getLocationByID -> getLocationById

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+  /services:
+    get:
+      tags:
+       - services
+      summary: Get services
+      description: 'Gets all `Service` objects. Can be filtered by parameters.'
+      operationId: getServicesList
+      parameters:
+        - in: query
+          name: filter[query]
```
tsoliangwu0130: Should this be `search`?

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+        - $ref: "#/components/parameters/isOpen"
+        - $ref: '#/components/parameters/pageNumber'
+        - $ref: '#/components/parameters/pageSize'
+      responses:
+        '200':
+          description: Successful response
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ServicesResultObjects'
```
tsoliangwu0130: ServicesResultObjects -> ServiceSetResult. I just realized we only add naming convention on [style guide](https://github.com/osu-mist/style-guides/blob/master/OpenAPI.md#result-object--resource-object) but forgot to update our skeleton (openapi3 branch). Can you create a PR for the fix?

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
-      operationId: getPetById
+        - services
+      description: 'Finds service by ID'
+      operationId: getServiceById
       parameters:
-        - name: id
+        - name: serviceId
           in: path
-          description: Unique Pet ID, UUID format
+          description: 'ID of location to be searched by'
```
tsoliangwu0130: ID of service.

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+            type: string
       responses:
         '200':
           description: Successful response
-          schema:
-            $ref: '#/definitions/PetResult'
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ServicesResultObject'
```
tsoliangwu0130: ServicesResultObject -> ServiceResult.

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
-        '409':
-          description: Conflict
+            type: string
+      responses:
+        '200':
+          description: Successful response
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ServicesResultObjects'
```
tsoliangwu0130: ServicesResultObjects -> ServiceSetResult

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+        - in: query
+          name: filter[query]
+          description: 'Query to search through buildings. Looks through the building name, abbreviation.'
+          required: false
           schema:
-            $ref: '#/definitions/ErrorResult'
+            type: string
+        - $ref: "#/components/parameters/isOpen"
+        - $ref: '#/components/parameters/pageNumber'
+        - $ref: '#/components/parameters/pageSize'
```
tsoliangwu0130, line 247: I'm not sure, but we have `include` compond parameter in `/locations` endpoint to return related services, should we also have one for `/services` to return related locations as well?

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
-        format: uri
-        description: The first page of data
-      last:
+          $ref: "#/components/responses/500"
+components:
+  parameters:
+    pageSize:
+      in: query
+      name: page[size]
+      description: 'Number of results to return per page. Used in pagination. Default value is 10.'
```
tsoliangwu0130: Your default value is actually set to 25 😛 You can just copy the entire pagination parameters from [here](https://github.com/osu-mist/express-api-skeleton/blob/966cdd24ee886005d6c4d6c081546712a2a4dfe7/openapi.yaml#L129-L147)

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+        application/json:
+          schema:
+            $ref: '#/components/schemas/ErrorResult'
+    '404':
+      description: Not found
+      content:
+        application/json:
+          schema:
+            $ref: '#/components/schemas/ErrorResult'
+    '409':
```
tsoliangwu0130: Do we need 409 anywhere?

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+          $ref: '#/components/schemas/Links'
+        data:
+          $ref: '#/components/schemas/LocationsResourceObject'
+        included:
+          type: array
+          items:
+            $ref: '#/components/schemas/ServicesResourceObject'
+    LocationSetResult:
+      properties:
+        links:
```
tsoliangwu0130, line 426: You also need a `meta` object for pagination details: https://github.com/osu-mist/express-api-skeleton/blob/openapi3/openapi.yaml#L174-L191

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+    LocationResult:
+      properties:
+        links:
+          $ref: '#/components/schemas/Links'
+        data:
+          $ref: '#/components/schemas/LocationsResourceObject'
+        included:
+          type: array
+          items:
+            $ref: '#/components/schemas/ServicesResourceObject'
```
tsoliangwu0130: ServicesResourceObject -> ServiceResouce

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+        application/json:
+          schema:
+            $ref: '#/components/schemas/ErrorResult'
+  schemas:
+    LocationResult:
+      properties:
+        links:
+          $ref: '#/components/schemas/Links'
+        data:
+          $ref: '#/components/schemas/LocationsResourceObject'
```
tsoliangwu0130: LocationsResourceObject -> LocationResource

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+            $ref: '#/components/schemas/LocationsResourceObject'
+        included:
+          type: array
+          items:
+            $ref: '#/components/schemas/ServicesIncluded'
+    ServicesResultObject:
+      properties:
+        links:
+          $ref: '#/components/schemas/Links'
+        data:
```
tsoliangwu0130, line 442: Should we have included here?

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+          $ref: '#/components/schemas/ServicesResourceObject'
+    ServicesIncluded:
```
tsoliangwu0130: Base on [JSONAPI](https://jsonapi.org/format/#document-top-level), `included` should be

> an array of resource objects that are related to the primary data and/or each other (“included resources”).

So why we need a new schema for this instead of using ServiceResource?

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+          type: string
+        attributes:
+          $ref: '#/components/schemas/Attributes'
+        links:
+          $ref: '#/components/schemas/SelfLink'
+        relationships:
+          $ref: '#/components/schemas/RelationshipsInLocations'
+    ServicesResultObjects:
+      properties:
+        links:
```
tsoliangwu0130, line 496: You need a `meta` object for this one as well.

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+    ServicesResultObject:
+      properties:
+        links:
+          $ref: '#/components/schemas/Links'
+        data:
+          $ref: '#/components/schemas/ServicesResourceObject'
+    ServicesIncluded:
+      properties:
+        id:
+          type: string
```
tsoliangwu0130, line 453: Can you add description and example for all of these fields? Example: https://github.com/osu-mist/persons-api/blob/v2/openapi.yaml#L846-L849

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+        type:
+          type: string
```
tsoliangwu0130: What's the value of type? https://github.com/osu-mist/persons-api/blob/v2/openapi.yaml#L850-L851

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+        links:
+          $ref: '#/components/schemas/ServiceLinks'
+    LocationsResourceObject:
+      properties:
+        id:
+          type: string
+        type:
+          type: string
+        attributes:
+          $ref: '#/components/schemas/Attributes'
```
tsoliangwu0130: Did you reuse this object anywhere else? If not, I think you can directly define your attributes here. If you want to define them in other places for the clearness, that's also fine, but you should consider renaming it to LocationAttributes for the consistency.

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+    ServicesResourceObject:
+      properties:
+        id:
+          type: string
+        type:
+          type: string
+        attributes:
+          $ref: '#/components/schemas/ServicesAttributes'
+        links:
+          $ref: '#/components/schemas/ServiceLinks'
```
tsoliangwu0130: What's difference between ServiceLinks and SelfLink?

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+        relationships:
+          $ref: '#/components/schemas/RelationshipsInServices'
+    SelfLink:
+      properties:
+        self:
+          type: string
+          description: Url to this resource / location.
```
tsoliangwu0130: Unless we missed any filed on the skeleton, I recommend directly use [whatever we've defined there](https://github.com/osu-mist/express-api-skeleton/blob/openapi3/openapi.yaml#L192-L197) instead of defining yourself again. By doing that, we can make sure all of our APIs follow the same style/standard and it will also be easier for us if we need to fix a bug.

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+    Attributes:
+      properties:
+        name:
+          type: string
+          description: Name of the location.
```
tsoliangwu0130, line 551: Can you add `description`, `example`, `format` (if applied), and `nullable` fields for all of these attributes? Example: https://github.com/osu-mist/persons-api/blob/v2/openapi.yaml#L891-L920

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+        abbreviation:
```
tsoliangwu0130: As mentioned, it should be split to two: banner and arcGis.

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+          example: Corvallis
+        state:
+          type: string
+          example: OR
+        zip:
+          type: string
+          example: '97330'
+        telephone:
+          type: string
+          example: 541-111-1111
```
tsoliangwu0130: Phone numbers need to be quoted.

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+            format: url
+            example: https://example.oregonstate.edu/image.jpg
+        images:
+          type: array
+          items:
             type: string
-            description: Name of owner
-            example: Harry Potter
-          species:
+            format: url
```
tsoliangwu0130: You should use `uri` instead of `url`. Also, I recommend rename this attribute to `imageUris` or `imageLinks` or something similar to make it clear.

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
-      detail:
-        type: string
-        description: A long description of the error that may contain instance-specific details
-      links:
-        properties:
-          about:
+            example: College of Engineering
+        website:
+          type: string
+          format: url
```
tsoliangwu0130: uri.

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+          example: https://example.oregonstate.edu
+        sqft:
+          type: string
```
tsoliangwu0130: Should this be number?

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+        calendar:
+          type: string
```
tsoliangwu0130, line 682: What is it? An uri?

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+        campus:
+          type: string
+          example: Corvallis
```
tsoliangwu0130: Your parameter doesn't have this option. We should make them be consistent (case sensitive). You also need an enum here.

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+        type:
+          type: string
+          enum: [building, dining, other, parking, cultural-center]
+          example: building
```
tsoliangwu0130: If you have enum, no need to have example, same as boolean.

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+          description: >
+            If true, access to this location's restrooms may be limited.
+            A residence hall might restrict bathroom use or building access to
+            residents only.
+          type: boolean
+        giRestroomLocations:
+          description: >
+            Room numbers of gender inclusive restrooms within location.
+            Some restrooms may not be ADA compliant.
+          type: string
```
tsoliangwu0130: I think it should be an array of strings.

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
-    properties:
-      errors:
-        type: array
-        items:
-          $ref: '#/definitions/ErrorObject'
+            example: recreation
+        openHours:
+          type: array
+          items:
+            $ref: '#/components/schemas/DayHours'
```
tsoliangwu0130: Array of DayHours?

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+          type: string
+        type:
+          type: string
+    DayHours:
+      description: >
+        This object represents open hours for a location, relative to the date
+        that the API returns the response. 1 represents the current day,
+        2 represents the next day, etc. All times are in UTC.
+      properties:
+        1:
```
tsoliangwu0130: This should be quoted.

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+            $ref: '#/components/schemas/TimeSlot'
+    TimeSlot:
+      properties:
+        start:
+          type: string
+          format: date-time
+        end:
+          type: string
+          format: date-time
+    Links:
```
tsoliangwu0130: Try naming variables/fields/functions more descriptive. Also, you should directly use the definition from here: https://github.com/osu-mist/express-api-skeleton/blob/openapi3/openapi.yaml#L198

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+          type: string
+          description: A long description of the error that may contain instance-specific
+            details
+        links:
+          properties:
+            about:
+              type: string
+              format: uri
+              description: A link to further information about the error
+              example: https://developer.oregonstate.edu/documentation/error-reference#1234
```
tsoliangwu0130: A new line needed EOF.

tsoliangwu0130 (changes requested):

openapi.yaml
tsoliangwu0130: Sure, I agree that it's helpful that having a search parameter to search across fields. But I also think we still need parameters for specific search. I'm open to any discussion, we can discuss about it in person or bring this up in our team meeting if you want.

openapi.yaml
Ruefa: This doesn't seem that uncommon to do. Just quality of life for our users.

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+                - B2
+                - B3
+                - C
+                - R
+        - in: query
+          name: filter[adaParkingSpaceCount][gte]
+          description: 'Only include locations with ADA parking space greater than or equal to this amount'
+          required: false
+          schema:
+            type: integer
```
Ruefa: I feel these should be
```
type: number
format: integer
```
for consistency

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+        - $ref: '#/components/parameters/pageNumber'
+        - $ref: '#/components/parameters/pageSize'
+        - in: query
+          name: include
+          required: false
+          schema:
+            type: array
+            items:
+              type: string
+              enum: [services]
```
Ruefa: Seems odd to have an array with only 1 enum unless it will have more in the future. Probably should be a boolean otherwise.

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+          $ref: '#/components/responses/500'
+  /locations/{locationId}/services:
+    get:
+      tags:
+        - locations
+        - services
+      description: 'Lists services related to a location'
+      operationId: getServices
+      parameters:
+        - name: locationId
```
Ruefa: This parameter is repeated. Should be moved to `components/parameters`

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+      operationId: getServicesList
+      parameters:
+        - in: query
+          name: filter[query]
+          description: 'Query to search through buildings. Looks through the building name, abbreviation.'
+          required: false
           schema:
-            $ref: '#/definitions/ErrorResult'
+            type: string
+        - $ref: "#/components/parameters/isOpen"
```
Ruefa: should be single quotes

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+        - $ref: '#/components/parameters/pageSize'
+      responses:
+        '200':
+          description: Successful response
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ServicesResultObjects'
+        '400':
+          $ref: "#/components/responses/400"
```
Ruefa: Should be single quotes. Can you change this everywhere?

openapi.yaml
Ruefa: I think this enum is repeated in a parameter somewhere. You can use pointers here.

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+                The default distance unit is mile '
+          type: number
+          format: float
+          example: 1.8
+        tags:
+          description: 'Tags associated with this location'
+          type: array
+          items:
+            type: string
+            example: cultural-center
```
Ruefa: Could this be an enum?

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+        parkingZoneGroup:
+          description: 'Zonegroup of location if location is parking'
+          type: string
+          example: A1
```
Ruefa: I think this can be an enum

openapi.yaml
```diff
@@ -1,279 +1,678 @@
...
+        evParkingSpaceCount:
+          description: 'Count of electric vehicle parking spaces if the location is parking'
+          type: integer
```
Ruefa: Maybe should use
```
type: number
format: integer
```

Ruefa (changes requested):

openapi.yaml
jemisonf: Cool, that sounds good.

openapi.yaml
jemisonf: Thanks, was just wondering about this.

openapi.yaml
jemisonf: Is it an issue to have two paths with the same operationId? Reading the [openapi docs](https://swagger.io/docs/specification/paths-and-operations/) it sounds like it might be intended to be unique on a per-path basis.

tsoliangwu0130: @jemisonf right, operation ID should be unique per path. I think this path should use getServices since it's the main services endpoint. You can rename the operation ID to something specific for the relation endpoint.

jemisonf: Cool. I'll use `getLocationServices`.

openapi.yaml
jemisonf: I think that if we're using multiple abbreviations the style guide doesn't necessarily make more sense than what we have now. If we were change the endpoint to use a single abbreviation per unit then it would make sense to switch. @Ruefa what do you think about that since you were supportive of keeping multiple abbreviations?

openapi.yaml
jemisonf: Makes sense :+1:

openapi.yaml
jemisonf: This is the convention I've seen with the `included` parameter in other APIs, which I think is nice for consistency's sake. Not super attached to this approach though.

openapi.yaml
jemisonf: It seems more practical to me that you'd want to get a location bundled with its services than service bundled with its location, so I'm not sure it would be worth it to implement. Thoughts @tsoliangwu0130?

openapi.yaml
jemisonf: Good catch :smile: I'll fix that

openapi.yaml
jemisonf: Looks like no, I'll take it out

openapi.yaml
tsoliangwu0130: Good questions. Since we currently do have relationship in service objects in v1, I think some people may want to know related locations when querying services. I'm not 100% sure, but I imagine the logic should be pretty similar between these two cases so ideally it shouldn't be too hard. I'm open on this though, so maybe we can get feedback from others on this. @Ruefa @kalcicd thoughts?

openapi.yaml
jemisonf: So then we'd have three filters -- `filter[search]`, `filter[location]` and `filter[name]`? That works for me.

openapi.yaml
Ruefa: I suppose that sometimes `included` will need to be a enum array so to be consistent it should always be.

jemisonf:
Some ideas for how to structure the `DayHours` representation:

First: change the `"1"`, `"2"`, etc keys to `"monday"`, `"tuesday"`, etc, making sure to be clear in the documentation that the times start from the current day and show the schedule for the next week. This seems more human-readable, but whether or not it's easier for developers seems dependent on exactly how their system is set up.

Second: let developers request a date range they want open times for (depending on if we have historical/future data for this), and return the times for that specific date range, with the "keys" for each time being relevant date. At this point it would probably be worth structuring `DayHours` as its own resource, even if it's not on the backend, to better take advantage of json:api semantics.

I will say that the current design isn't super elegant, but if it works it works, and I'd want to be cautious about creating something that's prettier but ends up making our clients' lives harder. I am curious if there are any specific client needs that aren't being met by the current design.

@tsoliangwu0130 let me know what you think.

openapi.yaml
```diff
@@ -1,279 +1,720 @@
...
+            - `m` - meters
+            - `cm` - centimeters
+            - `mm` - millimeters
+            - `nmi` - nautical miles
+
+            Should be used with the `filter[distance]` parameter
+          required: false
+          schema:
+            type: string
+            enum: [mi, yd, ft, in, km, m, cm, mm, nmi]
```
jemisonf, line 85: here's the solution for distance unit abbreviations. Looks much nicer in swagger UI right now.

openapi.yaml
```diff
@@ -1,279 +1,720 @@
...
           required: false
-          type: string
+          schema:
+            type: string
+        - in: query
+          name: filter[name]
+          description: Search buildings by name
+          required: false
+          schema:
+            type: string
```
jemisonf: added a `filter[name]` query to search specifically for buildings by name instead of combined name and abbreviation in `search`

openapi.yaml
jemisonf: It's slightly different -- we need `attributes` and `id` in the included document, but we also don't want to include the `relationships` key. So it's neither the same as `ServicesResourceObject` or `ServiceResult`. I think we _could_ include `relationships`, but I'm not sure it's practical.

openapi.yaml
```diff
@@ -1,279 +1,720 @@
...
+                type: array
+                items:
+                  type: array
+                  items:
+                    type: number
+                    format: float
+        summary:
+          type: string
+        description:
+          type: string
```
jemisonf: noticing that `summary` and `description` do not have super clear functions in the example data I have. Same for `calendar`, which I wasn't able to find _any_ values for. I'm wondering if anyone can provide guidance about what these three attributes are supposed to be used for.

jemisonf: Changing name of summary to zoneSummary

jemisonf: Label description as something like "miscellaneous information about the building", and combine description and descriptionHTML into a single object.

jemisonf: `calendar` is a URL to a calendar file for the building, but we're having trouble verifying that

openapi.yaml
jemisonf: Adding `filter[name][fuzzy]` to allow fuzzy searching of a location's name

jemisonf:
@tsoliangwu0130's idea for how to implement `DayHours`: use an array of objects instead of an object with keys "1".."7", and give each object a timestamp to show which day it represents.

openapi.yaml
```diff
@@ -1,279 +1,739 @@
...
+      items:
+        $ref: '#/components/schemas/TimeSlot'
+    TimeSlot:
+      properties:
+        start:
+          type: string
+          format: date-time
+        end:
+          type: string
+          format: date-time
```
jemisonf: Here's the new `DayHours`. I didn't  add an additional timestamp for the days because the `start` and `end` values are already in timestamp format.

openapi.yaml
tsoliangwu0130: I think you missed this comment.

openapi.yaml
tsoliangwu0130: I guess you missed this one too.

openapi.yaml
tsoliangwu0130: I think you missed this one.

openapi.yaml
```diff
@@ -1,279 +1,747 @@
...
+          type: integer
+          description: Number of results per page
+          example: 25
+    SelfLink:
+      properties:
+        self:
+          type: string
+          description: Url to this resource / location.
+    LocationsAttributes:
+      properties:
```
tsoliangwu0130, line 547: If you preview this openapi file by using swagger editor, you will find the example response body will look like this:

```json
{
  "links": {
    "first": "string",
    "last": "string",
    "prev": "string",
    "next": "string"
  },
  "data": [
    {
      "id": "5c4f1efcbaada0c63e",
      "type": "service",
      "attributes": {
        "name": "string",
        "abbreviation": "string",
        "latitude": 0,
        "longitude": 0,
        "geometry": {
          "type": "Polygon",
          "coordinates": [
            [
              [
                0
              ]
            ]
          ]
        },
        "zoneSummary": "string",
        "description": {
          "html": "string",
          "text": "string"
        },
        "address": "12345 NE STREET RD",
        "city": "Corvallis",
        "state": "OR",
        "zip": "97330",
        "telephone": "541-111-1111",
        "fax": "541-111-1111",
        "thumbnails": [
          "https://example.oregonstate.edu/image.jpg"
        ],
        "images": [
          "https://example.oregonstate.edu/image.jpg"
        ],
        "departments": [
          "College of Engineering"
        ],
        "website": "https://example.oregonstate.edu",
        "sqft": "string",
        "calendar": "string",
        "campus": "corvallis",
        "type": "building",
        "giRestroomCount": 0,
        "giRestroomLimit": true,
        "giRestroomLocations": [
          "0111A"
        ],
        "synonyms": [
          "Gymnasium"
        ],
        "openHours": [
          {
            "start": "2020-05-06T17:32:34.067Z",
            "end": "2020-05-06T17:32:34.067Z"
          }
        ],
        "distance": 1.8,
        "tags": [
          "building"
        ],
        "parkingZoneGroup": "A1",
        "evParkingSpaceCount": 0,
        "adaParkingSpaceCount": 0,
        "motorcycleParkingSpaceCount": 0,
        "weeklyMenu": "https://app.uhds.oregonstate.edu/api/dining/weeklymenu/index?loc=1"
      },
      "links": {
        "self": "string"
      },
      "relationships": {
        "services": {
          "data": [
            {
              "id": "string",
              "type": "string"
            }
          ],
          "links": {
            "self": "string"
          }
        }
      }
    }
  ],
  "included": [
    {
      "id": "string",
      "type": "string",
      "attributes": {
        "name": "string",
        "tags": [
          "recreation"
        ],
        "openHours": [
          {
            "start": "2020-05-06T17:32:34.067Z",
            "end": "2020-05-06T17:32:34.067Z"
          }
        ]
      },
      "links": {
        "self": "string"
      }
    }
  ],
  "meta": {
    "totalResults": 10,
    "totalPages": 10,
    "currentPageNumber": 1,
    "currentPageSize": 25
  }
}
```

As you can see, many of these properties' values are "string" which is not good. As I mentioned before, we need to provide "example" to all of them so that our clients can easily understand when they read this doc.

openapi.yaml
```diff
@@ -1,279 +1,747 @@
...
+            - `nmi` - nautical miles
+
+            Should be used with the `filter[distance]` parameter
+          required: false
+          schema:
+            type: string
+            enum: [mi, yd, ft, in, km, m, cm, mm, nmi]
+        - $ref: '#/components/parameters/isOpen'
+        - in: query
+          name: filter[campus][oneOf]
```
tsoliangwu0130, line 88: The swagger doc looks correct, but I just realized if you preview this in the swagger editor, it will only allow you to search "one" campus. Same for all other `oneOf` filter. Thoughts?

openapi.yaml
```diff
@@ -1,279 +1,747 @@
...
           required: false
-          type: string
+          schema:
+            type: string
+        - in: query
+          name: filter[name]
+          description: Search for a building with this exact name
+          required: false
+          schema:
+            type: string
```
tsoliangwu0130, line 42: Can you also add examples to all of these parameters so that our clients can see a placeholder value when they read this doc via our developer portal?

tsoliangwu0130 (changes requested):
I believe you missed most of my last comments in my last review. So I'm going pause here until you address them :)

openapi.yaml
jemisonf: It definitely _should_ allow you to select multiple values. I'm actually not sure what the solution to this is -- the specification looks correct to me, but it's definitely not working. I noticed that the selected value doesn't show up at all in the generated query, which makes me wonder if it's an issue with Swagger UI. 

Looking for other examples, I found this on line 529 of the persons v2 openapi doc:
```yaml
        - name: filter[emailType]
          in: query
          required: false
          description: |
            Type of email
            - `ONID` - Student
            - `STU` - Non-directory Student
            - `VEND` - Vendor
            - `VWEB` - Vendor
            - `EMPL` - Employee
            - `VPAY` - Vendor Pay
            - `EMER` - Emergency Contact Persons Email for Student
          schema:
            type: array
            items:
              type: string
              enum: 
                &emailTypeEnums [ONID, STU, VEND, VWEB, EMPL, VPAY, EMER]
```

Which looks like it has the same issue. Any ideas about what the problem could be? I'm kind of stumped.

openapi.yaml
```diff
@@ -1,279 +1,753 @@
...
+            example: DxRC
+        - in: query
+          name: filter[arcGisAbbreviation]
+          description: Search by ArcGIS abbreviat ion
+          required: false
+          schema:
+            type: string
+            example: DxRC
```
jemisonf: I think these in theory can be different -- should they have different examples?

openapi.yaml
jemisonf: I'm not sure how to determine which fields are nullable -- is there a good source of truth for that information?

openapi.yaml
jemisonf: I couldn't find any examples, and all of the values in the `locations-combined.json` file are just empty strings, so I'm not sure.

jemisonf: Actually, looks like they're all `null`, so I'll update it to a number.

openapi.yaml
tsoliangwu0130: Most of them are the same so I think it's fine to use the same example here for now.

openapi.yaml
tsoliangwu0130: Nevermind, I think they actually allow you to select multiple values if you click the option with the `command` key (on Mac).

openapi.yaml
```diff
@@ -1,279 +1,795 @@
...
+          type: string
+          format: uri
+          nullable: true
+          description: The previous page of data
+        next:
+          type: string
+          format: uri
+          nullable: true
+          description: The next page of data
+    ServicePaginationLinks:
```
tsoliangwu0130: Why it is pagination links? What's the difference between this definition and `SelfLink`?

openapi.yaml
```diff
@@ -1,279 +1,795 @@
...
+          example: 5c4f1efcbaada0c63e
+        type:
+          type: string
+          enum: [service]
+    DayHours:
+      description: >-
+        This array represents open hours for a location, relative to the date
+        that the API returns the response. The first element represents the current day,
+        the second represents the next day etc. All times are in UTC.
+      type: array
```
tsoliangwu0130, line 850: You probably want to mention we will only show a week of schedule from now on in the description field and also set the length of this array to 7 to make sure clients won't run into index out of range error. See `Array Length` in https://swagger.io/docs/specification/data-models/data-types/

openapi.yaml
```diff
@@ -1,279 +1,795 @@
...
-        properties:
-          about:
+            example: College of Engineering
+        website:
+          type: string
+          description: the website of the location
+          format: uri
+          example: https://example.oregonstate.edu
+        sqft:
+          type: number
```
tsoliangwu0130, line 676: float or integer?

openapi.yaml
```diff
@@ -1,279 +1,795 @@
...
+            - R
+        evParkingSpaceCount:
+          description: Count of electric vehicle parking spaces if the location is parking
+          type: number
+          format: integer
+        adaParkingSpaceCount:
+          description: Count of ADA parking spaces if the location is parking
+          type: number
+          format: integer
+        motorcycleParkingSpaceCount:
```
tsoliangwu0130: I recommend merging these to a `spaceCount` object to make them easier to read.

openapi.yaml
```diff
@@ -1,279 +1,795 @@
...
+          description: Number of gender inclusive restrooms available in the location.
+          type: number
+          format: integer
+        giRestroomLimit:
+          description: >-
+            If true, access to this location's restrooms may be limited.
+            A residence hall might restrict bathroom use or building access to
+            residents only.
+          type: boolean
+        giRestroomLocations:
```
tsoliangwu0130: Same, what do you think if we merge all giRestroom attributes into one `giRestroom` object like:

```json
"giRestroom": {
  "limit": ...,
  "count": ...,
  "locations": ...
}
```

openapi.yaml
```diff
@@ -1,279 +1,795 @@
...
+      properties:
+        name:
+          type: string
+          description: Name of the location.
+          example: Dixon Recreation Center
+        bannerAbbreviation:
+          type: string
+          description: Abbreviation for the location
+          example: DxRC
+        arcGisAbbreviation:
```
tsoliangwu0130: Same thing, should we merge them to an `abbreviations` object with `banner` and `arcGis` attributes?

openapi.yaml
```diff
@@ -1,279 +1,795 @@
...
+          type: array
+          description: an array of URLs for thumbnail images of the location
+          items:
             type: string
-            description: Name of pet
-            example: Hedwig
-          owner:
+            format: uri
+            example: https://example.oregonstate.edu/image.jpg
+        imageUris:
```
tsoliangwu0130: Not sure can we merge these uris. Thoughts?

openapi.yaml
```diff
@@ -1,279 +1,795 @@
...
+          $ref: '#/components/schemas/SelfLink'
+    LocationsResourceObjectIdentifier:
+      properties:
+        id:
+          type: string
+          example: 5c4f1efcbaada0c63e
+        type:
+          type: string
+          enum: [location]
+    ServicesResourceObjectIdentifier:
```
tsoliangwu0130: We don't need to define these identifiers in separate objects if they are only used once.

openapi.yaml
```diff
@@ -1,279 +1,795 @@
...
+        id:
+          type: string
+        type:
+          type: string
+          enum: [service]
+        attributes:
+          $ref: '#/components/schemas/ServicesAttributes'
+        links:
+          $ref: '#/components/schemas/ServicePaginationLinks'
+    LocationsResourceObject:
```
tsoliangwu0130: You still miss this one 😛 

LocationsResourceObject -> LocationResource
ServicesResourceObject -> ServiceResource

tsoliangwu0130 (changes requested):

openapi.yaml
jemisonf: Not sure how this happened, it honestly might have been a find-and-replace error. I'm going to switch over to using `SelfLink`.

openapi.yaml
jemisonf: I'm _assuming_ integer, but I couldn't find any examples for this value, so I'm not 100% sure.

tsoliangwu0130: Yep let's use integer. Just wanna make everything clear on this doc.

openapi.yaml
jemisonf: I like this as well, I think it's a good approach.

openapi.yaml
jemisonf: Went for this:
```yaml
images:
          type: object
          description: full size and thumbnail image URLs for the location
          properties:
            fullSize:
              description: an array of URLs for images of the location
              type: array
              items:
                type: string
                format: uri
                example: https://example.oregonstate.edu/image.jpg
            thumbnails:
              type: array
              description: an array of URLs for thumbnail images of the location
              items:
                type: string
                format: uri
                example: https://example.oregonstate.edu/image.jpg
```

Let me know what you think.

openapi.yaml
```diff
@@ -351,11 +351,11 @@ components:
         links:
           $ref: '#/components/schemas/PaginationLinks'
         data:
-          $ref: '#/components/schemas/LocationsResourceObject'
+          $ref: '#/components/schemas/LocationsResource'
         included:
           type: array
           items:
-            $ref: '#/components/schemas/ServicesResourceObject'
+            $ref: '#/components/schemas/ServicesResource'
```
tsoliangwu0130: It's a singular `Service`. Same as `Location`.

tsoliangwu0130 (changes requested):

openapi.yaml
Ruefa: ¿Por que no los dos? I can think of use cases for both scenarios.

openapi.yaml
Ruefa: Not really. I did a lot `select _ from TABLE if _ is null`

openapi.yaml
jemisonf: Sounds like the consensus is we should do it -- I'll get it added in.

openapi.yaml
jemisonf: Hmm, @tsoliangwu0130 do you know?

tsoliangwu0130: There's no good way to tell for this API. The only way we can tell is to look at the source file I sent you or from AWS ES instance.

openapi.yaml
```diff
@@ -1,279 +1,858 @@
...
+        - in: query
+          name: filter[bannerAbbreviation]
+          description: Search by banner abbreviation
+          required: false
+          schema:
+            type: string
+            example: DxRC
+        - in: query
+          name: filter[arcGisAbbreviation]
+          description: Search by ArcGIS abbreviat ion
```
kalcicd: typ o

openapi.yaml
```diff
@@ -1,279 +1,858 @@
...
+          required: false
+          schema:
+            type: string
+            example: DxRC
+        - $ref: '#/components/parameters/pageNumber'
+        - $ref: '#/components/parameters/pageSize'
+        - in: query
+          name: include
```
kalcicd: This query param needs a description. Something simple like `Include related services` should do.

openapi.yaml
kalcicd: On that note, should /locations/:id have an include parameter? That could save the user from making two calls to /locations/:id and /locations/:id/services

kalcicd (changes requested):

jemisonf:
> On that note, should /locations/:id have an include parameter? That could save the user from making two calls to /locations/:id and /locations/:id/services

Couldn't find the original thread to reply this, but I agree this is a good idea. I think the default as is would be to just always include services, but it makes just as much sense that a user might want to _not_ include them.

openapi.yaml
```diff
@@ -1,279 +1,864 @@
...
+            format: integer
+        - in: query
+          name: filter[motoParkingSpaceCount][gte]
+          description: Only include locations with motorcycle parking space greater than or equal to this amount
+          required: false
+          schema:
+            type: number
+            format: integer
+        - in: query
+          name: filter[evParkingSpaceCount][gte]
```
tsoliangwu0130, line 155: These space count filters need examples.

openapi.yaml
```diff
@@ -1,279 +1,864 @@
...
+          $ref: '#/components/schemas/SelfLink'
+    LocationResource:
+      properties:
+        id:
+          type: string
+          example: 5c4f1efcbaada0c63e
+          description: ID of a location object
+        type:
+          type: string
+          enum: [service]
```
tsoliangwu0130: Should be `location`.

openapi.yaml
```diff
@@ -1,279 +1,864 @@
...
+        attributes:
+          $ref: '#/components/schemas/ServicesAttributes'
+        links:
+          $ref: '#/components/schemas/SelfLink'
+    LocationIncluded:
+      properties:
+        id:
+          type: string
+          example: 5c4f1efcbaada0c63e
+          description: ID of a location object
```
tsoliangwu0130: I think it should be a service?

openapi.yaml
```diff
@@ -1,279 +1,864 @@
...
+            $ref: '#/components/schemas/LocationIncluded'
+    ServiceResource:
+      properties:
+        id:
+          type: string
+          example: 5c4f1efcbaada0c63e
+          description: ID of a service object
+        type:
+          type: string
+          enum: [service]
```
tsoliangwu0130: `[service]` is used many times and I think it should be reusable. Example: https://github.com/osu-mist/persons-api/blob/v2/openapi.yaml#L703-L705

Same thing for `[location]`.

tsoliangwu0130 (changes requested):

openapi.yaml
```diff
@@ -1,279 +1,869 @@
...
+              description: Array of ring arrays for each physical structure of a location
+              type: array
+              items:
+                description: Array of points that make up a ring.
+                type: array
+                items:
+                  type: array
+                  items:
+                    type: number
+                    format: float
```
tsoliangwu0130: Can you also add example to this? It's now lookins as this in Swagger UI:

```json
"coordinates": [
  [
    [
      0
    ]
  ]
]
```

tsoliangwu0130: Also, as we discussed in person, each corrdinate should be an object.

tsoliangwu0130 (changes requested):

openapi.yaml
```diff
@@ -1,279 +1,914 @@
...
           schema:
-            $ref: '#/definitions/ErrorResult'
+            type: string
+      responses:
+        '200':
+          description: Successful response
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ServiceLocationSetResult'
```
jemisonf, line 303: I thought it made sense for this endpoint to return a response without included or meta. 
* Leaving out included because we're only focused on the relationship between this service and its locations, not others services/locations
* Leaving out meta because this endpoint doesn't use pagination

I made the same change in the locations/:id/services endpoint. Open to discussion on this but it makes sense to me.

tsoliangwu0130 (approved):

openapi.yaml
tsoliangwu0130: Sounds good to me. @Ruefa @kalcicd since you also review this, any thoughts?

tsoliangwu0130:
PR should be merged to locations api instead of here. PR: https://github.com/osu-mist/locations-api/pull/70 Closing PR here.

